PR #1, opened 2018-03-21
Minor cleanup, prettier output

nebkor:
This is a really great little tool! I noticed some compiler warnings when I built it locally, and I thought the box construction characters could be slightly fancier, so there you go. I split the changes into several commits, and the only one that changes behavior is the last one, 1c80cbf, "Make printing slightly prettier." I'm not sure this PR is worth merging, just FYI. Also, I might change it a little more aggressively, like factor it into a lib crate and a bin crate, and possibly try to speed it up.


bootandy:
thanks - Most definitely worth merging.

PR #3, opened 2018-03-22
Broken into files, inode/dev changes from inodes branch taken

nebkor:
The functions in main.rs had been broken out into the utils module a few commits ago (0faf795), so bringing over your changes by hand was pretty straight-forward. The next thing was breaking all the cfg compile directive controlled stuff into its own submodule for utils.

I actually don't recommend just merging this into your repo, because the history is all over the place. I tried to chunk the changes so that if you wanted to, you could linearize the change, or just jump to the new layout in one opaque commit (by simply copying it over).


src/utils/platform.rs
```diff
@@ -1,22 +1,25 @@
...
+pub fn get_metadata(
+    d: &std::fs::DirEntry,
+    use_apparent_size: bool,
+) -> Option<(u64, Option<(u64, u64)>)> {
     use std::os::linux::fs::MetadataExt;
     match d.metadata().ok() {
         Some(md) => {
             let inode = Some((md.st_ino(), md.st_dev()));
-            if s {
+            if use_apparent_size {
```
bootandy: The variable name 's' kept the function def on one line. I'm not sure that this is a valid reason to keep a bad name like 's' but just letting you know why it was 's'.

I'm happy with either version.

src/utils/platform.rs
```diff
@@ -1,22 +1,25 @@
...
+#[cfg(not(any(target_os = "macos", target_os = "linux")))]
 pub fn get_block_size() -> u64 {
     1024
 }
 
-#[cfg(target_os = "macos")]
+#[cfg(any(target_os = "macos", target_os = "linux"))]
 pub fn get_block_size() -> u64 {
     512
 }
```
bootandy: I've discovered its a bit weirder than that. I'll have a commit to fix up the get_block_size() soon.

PR #6, opened 2018-04-16
Don't ignore Cargo.lock

Infinisil:
See https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

I didn't include my own `Cargo.lock` file since you might want to commit your own which might have different versions.

I'll be able to package this for [nixpkgs](https://github.com/NixOS/nixpkgs) as soon as the next version is released with a `Cargo.lock` file.


Infinisil:
https://github.com/NixOS/nixpkgs/pull/39123 :D

PR #23, opened 2018-11-08
Update README.md

chrisjohnson00:
Fixed a typo where it says `dust` is slower than `du`


bootandy:
No, when I last timed du vs dust. Du was quicker.  

However I'm not using threads yet. Integrate Rayon and things could change.

PR #39, opened 2019-11-21
Clippy

lespea:
Fixes a bunch of clippy warnings which includes:
  * Don't box the Nodes (not needed since it's inside a vec)
  * Reversing the node children no longer requires collecting everything into an intermediary vec

Let me know if some of these changes you don't want (specifically the collapsed if-then-else branches you may prefer the original... I just went for cleaning everything for a first pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -145,9 +145,15 @@ dependencies = [
  "ansi_term 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "assert_cli 0.5.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "clap 2.33.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


bootandy:
Hi, Firstly thanks for the pull request. 

I like most of your changes but there are 2 things I'd like to change:

The collapsed if-then-else branches - I would like to leave these as they were as it is easier to understand them (we could put them in to two functions if you'd prefer).

Using the 'either' crate. To me it has hidden something in the way that 'get_children_from_node' works. I'd rather be more explicit even if I have to do a `.map().collect()` [If there are better ways to do this please feel free to suggest them].

But I like the other changes: removing box, changing iter to intoIter & getting rid of a few Somes.

So if you could undo the if/else squashing and the either crate I'd happily accept this merge. 

thanks.

PR #48, opened 2019-12-27
Fix colors on windows

lespea:
* Initialize ansi_term if we're on windows
* Silence build warning on non-unix targets
* Update ansi-term which has windows specific fixes (and assert_cli while I'm at it)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,16 +9,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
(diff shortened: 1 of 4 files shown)


bootandy:
This broke build due to version upgrade of assert_cli.

I pulled your commit into this merge request. 

https://github.com/bootandy/dust/pull/49#partial-pull-merging

lespea:
Awesome thanks!

PR #59, opened 2020-01-21
Allow multiple --ignore-directory flags

Celti:


bootandy:
I wondered about adding this myself. 

It looks like it adding 'multiple' has caused it to eat the last parameter
So running

`target/debug/dust -X target src`

is interpreted as ignore target and src instead of 'ignore target and run on src'

There is probably a flag in Clap to resolve this. Then I will happily merge this.

Celti:
Ah, heck, I knew I messed _something_ up there! Pushed an amended commit.

bootandy:
Looks good, will merge when the build goes green

Issue #60, opened 2020-01-21
Tests on Linux are filesystem-dependent


Celti:
When running `cargo test` while putting together #59, it came to my attention that tests on Linux are filesystem-dependent. `test_dir2` is 12K on my XFS home filesystem or my tmpfs /tmp, whereas the tests expect 24K (perhaps from ext4?).

This should at minimum be documented, but I'd recommend copying the test files to `/tmp` and running them there, as the vast majority of Linux systems default to `/tmp` being a tmpfs.

bootandy:
That sounds like a good idea. 

I was not aware that this would be a problem but it makes sense. I'll look at copying everything to /tmp for the tests.

bootandy:
An alternative would be to determine what file system we are looking at and then adjust the tests accordingly. 

We could also run the tests as 'result == X or result == Y or result == Z' and let any of them pass. 

This needs some thought and it should be resolved. 

Windows users were also reporting inconsistent numbers on their machines.

Celti:
I think adding specific per-filesystem test results is doomed to failure from the outset — there are simply _too many_ choices, where do you draw the line as to adding a new filesystem? Whose job is it to then update _all_ those filesystem-specific results? What if one filesystem's success matches also another filesystem's failure?

We already have `#[cfg(target_os)]` for Mac and Linux, I think adding one more for Windows and then requiring that the tests be run on APFS, tmpfs, or NTFS respectively isn't an unreasonable technical burden on either side.

bootandy:
>I think adding specific per-filesystem test results is doomed to failure from the outset — there are simply too many choices,

Agreed. 

I am also considering changing the tests so they only look at the file names and ignore the files sizes. However I don't like this idea as we would then loose the ability to test a 'complete' output. 

I'm doing a refactor at the minute and I'm keeping testing in mind. 

I think your idea of using the  `/tmp` directory may be the best option if the CI runners allow it.

bootandy:
https://github.com/bootandy/dust/commit/1f120de168fb2eae951bd81bd7566839834cbd16

This commit copies the tests files to the /tmp/ directory this should fix it.

PR #62, opened 2020-01-22
Fix windows compilation and testing

rivy:
- adds GitHub-Actions CI workflow (includes arm-linux, x86-linux, macos, and windows builds/testing)
  - automatic deployment to Github Releases when commit is version tagged (see example at <https://github.com/rivy/rs.dust/releases/tag/v0.4.9>; note: fully automatically generated)
- improves path manipulation portability
- builds / tests successfully locally and on all CI hosts
- wip / needs further investigation
  - feasability of (currently disabled) symlink tests on "windows"
  - investigate and/or update tests to handle minor size variations in testing output for different "windows" hosts
    - some one byte differences between local and CI host file sizes; ? bug or expected host variation
  - `cargo tarpaulin` inability to handle test coverage (use alternate framework?)

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -265,23 +265,59 @@ mod tests {
     #[test]
     fn test_simplify_dir_rm_subdir() {
         let mut correct = HashSet::new();
-        correct.insert("a/b".to_string());
+        correct.insert(
+            ["a", "b"]
+                .iter()
+                .collect::<std::path::PathBuf>()
+                .to_string_lossy()
+                .to_string(),
+        );
         assert_eq!(simplify_dir_names(vec!["a/b", "a/b/c", "a/b/d/f"]), correct);
     }
 
     #[test]
     fn test_simplify_dir_duplicates() {
         let mut correct = HashSet::new();
-        correct.insert("a/b".to_string());
+        correct.insert(
+            ["a", "b"]
+                .iter()
+                .collect::<std::path::PathBuf>()
+                .to_string_lossy()
+                .to_string(),
+        );
         correct.insert("c".to_string());
-        assert_eq!(simplify_dir_names(vec!["a/b", "a/b//", "c", "c/"]), correct);
+        assert_eq!(
+            simplify_dir_names(vec![
+                "a/b",
+                "a/b//",
+                "a/././b///",
+                "c",
+                "c/",
+                "c/.",
+                "c/././",
+                "c/././."
+            ]),
+            correct
+        );
     }
     #[test]
     fn test_simplify_dir_rm_subdir_and_not_substrings() {
         let mut correct = HashSet::new();
         correct.insert("b".to_string());
-        correct.insert("c/a/b".to_string());
-        correct.insert("a/b".to_string());
+        correct.insert(
+            ["c", "a", "b"]
+                .iter()
+                .collect::<std::path::PathBuf>()
+                .to_string_lossy()
+                .to_string(),
+        );
+        correct.insert(
+            ["a", "b"]
+                .iter()
+                .collect::<std::path::PathBuf>()
+                .to_string_lossy()
+                .to_string(),
+        );
         assert_eq!(simplify_dir_names(vec!["a/b", "c/a/b/", "b"]), correct);
     }
 
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -38,8 +38,9 @@ impl PartialEq for Node {
 }
 
 pub fn is_a_parent_of(parent: &str, child: &str) -> bool {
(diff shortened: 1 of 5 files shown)


src/utils/mod.rs
```diff
@@ -38,16 +38,17 @@ impl PartialEq for Node {
 }
 
 pub fn is_a_parent_of(parent: &str, child: &str) -> bool {
-    (child.starts_with(parent) && child.chars().nth(parent.chars().count()) == Some('/'))
-        || parent == "/"
+    let path_parent = std::path::Path::new(parent);
+    let path_child = std::path::Path::new(child);
+    (path_child.starts_with(path_parent) && !path_parent.starts_with(path_child))
```
bootandy: I've been staring at this for sometime and I think it is good. 

Can you add a test for this case as it was bothering me:

`is_a_parent_of('/usr/folder', '/usr/folder_not_a_child')  == False` 

?
@rivy

bootandy:
>"windows" hosts    some one byte differences between local and CI host file sizes; ? bug or expected host variation

We also have a similar problem on linux if the filesystem is a different type we will see different numbers. https://github.com/bootandy/dust/issues/60. I'm not sure what the best way round this is. I'm considering running it once to see what the value should be and then dynamically generating the tests. Or we could add a statement like: pass if test=A or test=B.

bootandy:
Overall It looks good, I want to play with this branch myself for a bit but I think I'd be happy to merge it.

bootandy:
Although I am going to be offline for the next 10 days, so apologies in advance for going quiet.

src/utils/mod.rs
rivy: I tried to find a portable std::path-based method to determine parent-child relation and couldn't find anything. This was what I finally came up with as a possible design. But, yep, it's a bit twisty and I mentally bounced the code around for a while as well. I had to add a few test cases just to help prove it to myself. I'll happily add any extra test cases.

rivy: Done, and it passes (unlike `std::String::starts_with()`, `std::path::Path.starts_with()` says that it "only considers whole path components to match" which is why this works).

I'll add the commit later tonight or tomorrow.

rivy:
No worries on the timing.

The testing is breaking (solely based on sizes) on my local windows `wsl` ubuntu as well. I'll have a look at it over the next week or two and see if I can come up with anything that would make testing more robust.

I also cut some CI code that builds/deploys a deb package. It might satisfy issue #7. I'll add it back in as a separate commit.

I also don't see the GitHub-Actions CI building on your repo side. I'm not quite sure how to get it enabled, but I'll google around if it's not obvious from your end.

rivy:
I just looked through my repository settings and it looks like you'll need to enable Actions via <https://github.com/bootandy/dust/settings/actions> for the CICD action to run for the PR.

I have "Enable local and third party Actions for this repository" selected. I think that's the setting needed since the CI here pulls in actions to setup rust and cross-compile.

rivy:
Just an additional note here to prompt myself ...

I'm seeing "Did not have permissions for all directories" (though I surely have permissions for all searched directories) during execution on windows platforms. I'll investigate further.

bootandy:
I may be stating the obvious here but:

> "Did not have permissions for all directories" 

This message is printed to stderr if you don't have permission on 1 or more files or directories which dust encounters.

rivy:
> I may be stating the obvious here but:
> 
> > "Did not have permissions for all directories"
> 
> This message is printed to stderr if you don't have permission on 1 or more files or directories which dust encounters.

I found the problem. `winapi_util::Handle::from_path()` was being used to enumerate directory entries, but the handles returned for directories are useless. 

> "If you use `from_path()` on a directory, then subsequent queries using that handle will fail."
>
> ref: "Struct winapi_util::Handle" from <https://docs.rs/crate/winapi-util>

`winapi_util::Handle::from_path_any()` needs to be used when the path might be a file or directory.

The bug was fixed with one change at the first use. But now I'm looking at the second use, trying to figure out if a directory could ever be an argument. I would think that `from_path_any()` should be used in all cases, but I'm not sure if there's a performance penalty.

rivy:
I fixed/updated and polished a few things based upon this discussion.

Additionally, I've also created a commit that refactors the code to use `PathBuf` instead of `String`. I think it simplifies some things, making portable path handling easier, and avoids some of the conversion churn between strings and paths.

Here's the commit (based on top of this branch) ... <https://github.com/rivy/rs.dust/compare/add.gha...rivy:rf.pathbuf>.

If you think the change is useful, I'd like to add it to this PR. But I can open it in a new PR if, again, you'd like the change but prefer a new / separate PR.

PR #64, opened 2020-02-03
Refactor ~ use Path/PathBuf instead of &str/String

rivy:
This minimizes the string conversion/copy "fiddliness" and helps somewhat in code comprehension and maintaining/encouraging cross-platform compatibility. 

There may be some additional work to be done with references, especially in the containers, that might be useful to further minimize copying. But it's a bit beyond my understanding of the language at the moment to go beyond this point.

Let me know if this is helpful, and if so, whether you'd like any changes.

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -70,9 +70,9 @@ pub fn simplify_dir_names<P: AsRef<Path>>(filenames: Vec<P>) -> HashSet<PathBuf>
     top_level_names
 }
 
-pub fn get_dir_tree(
-    top_level_names: &HashSet<PathBuf>,
-    ignore_directories: Option<Vec<PathBuf>>,
+pub fn get_dir_tree<P: AsRef<Path>>(
+    top_level_names: &HashSet<P>,
+    ignore_directories: &Option<Vec<PathBuf>>,
     apparent_size: bool,
     limit_filesystem: bool,
     threads: Option<usize>,
@@ -90,7 +90,7 @@ pub fn get_dir_tree(
             b,
             apparent_size,
             &restricted_filesystems,
-            &ignore_directories,
+            ignore_directories,
             &mut data,
             &mut permissions,
             threads,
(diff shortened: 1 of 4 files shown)


bootandy:
The code is dealing with Paths here not Strings so it does make sense to use PathBuf throughout the codebase. I'll have a look at this later.

Thanks

src/utils/mod.rs
```diff
@@ -108,29 +109,26 @@ fn get_allowed_filesystems(top_level_names: &HashSet<String>) -> Option<HashSet<
...
     // 2. removing interior '.' ("current directory") path segments
     // 3. removing trailing extra separators and '.' ("current directory") path segments
     // * `Path.components()` does all the above work; ref: <https://doc.rust-lang.org/std/path/struct.Path.html#method.components>
     // 4. changing to os preferred separator (automatically done by recollecting components back into a PathBuf)
-    path.as_ref()
-        .components()
-        .collect::<std::path::PathBuf>()
-        .to_string_lossy()
-        .to_string()
+    path.as_ref().components().collect::<PathBuf>()
```
bootandy: do you think we can use canonicalize? https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.canonicalize

src/utils/mod.rs
```diff
@@ -236,18 +240,31 @@ pub fn find_big_ones(new_l: Vec<(String, u64)>, max_to_show: usize) -> Vec<(Stri
...
 
     for name in top_level_names {
-        let my_max_depth = name.matches(std::path::is_separator).count() + max_depth as usize;
-        let name_ref: &str = name.as_ref();
+        let my_max_depth = name
+            .components()
+            .filter(|&c| match c {
+                std::path::Component::Prefix(_) => false,
+                _ => true,
+            })
```
bootandy: can't we remove the above `filter` isn't name.components.count() enough?

src/utils/mod.rs
rivy: Yes that's a possible alternative, but this method always succeeds. `fs::canonicalize()`, on the other hand, requires an existing path or returns an error which would then have to be handled.

src/utils/mod.rs
rivy: No, paths can have some odd preliminary ("Prefix") bits (for example, from windows, "\\?\" or "\\UNC\") which need to be removed here to calculate the depth correctly.

src/utils/mod.rs
bootandy: good reasoning, agreed

bootandy:
Looks good, thanks for this.

Minor note: This change means the '-X' flag now works on directories. Before it worked on substrings of directory paths. Before '-X s' would ignore the 'src/' directory. Now it wont. I think this is fine, but I am logging it here incase someone asks.

PR #79, opened 2020-02-29
use LS_COLORS for largest subdirectories

NeelChotai:
Uses directory colour defined in $LS_COLORS to paint the largest subdirectories instead of red.

Fixed a compiler warning, updated the readme, changed args descriptions and made some consistency changes to clap args.


bootandy:
Looks good, I like your changes. 

We have a conflict as I moved the line handling colorized output just before this was opened, but this should be easy to fix when you rebase. 

I will happily merge when the conflict is resolved.

NeelChotai:
Thank you.

Rebased without conflicts, would you like me to squash everything in to one commit?

bootandy:
Thanks,

No. I prefer several small commits.

Thanks for introducing me to ls_colors library too. It is giving me ideas.

PR #81, opened 2020-03-10
Maint/CI ~ fix features option for GHA `cargo ...` and `cross ...`

rivy:
This fixes the failing empty features issue more generically and robustly.

I just noticed the GHA builds were failing on another repo and created this fix. I see that you've fixed it here by removing support for the features option completely. This implementation works whether the build uses features or not. It'll be more robust in the case that you add any features option at a later date.

Since, you're current implementation works, I understand if you don't want to merge this ... but I think that this is a better version which a future you might appreciate. 😄


bootandy:
Thanks, I'm sure this is a better thing to do. I was just 'doing the simplest thing to make the build work again'

PR #97, opened 2020-08-01
add brew command in README

pom421:


README.md
```diff
@@ -18,6 +18,10 @@ Because I want an easy way to see where my disk is being used.
 
 * `cargo install du-dust`
 
+#### Brew (Mac OS)
```
TGotwig, line 21: Maybe we could add a 🍺-Emoji 🙂

pom421: Almost necessary, indeed ;)

TGotwig: 😁🍻 Since it is a very small change, it would be good to squash them into one commit and do a force-push 🚀

PR #109, opened 2020-08-18
Support listing directories by number of files

bootandy:
Based of https://github.com/bootandy/dust/pull/104

Idea is to allow users to find the number of files in each directory
instead of size.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,8 @@ unicode-width = "0.1"
 ignore="0.4"
 crossbeam-channel = "0.4"
 walkdir="2.3"
+# todo use num_format instead
+thousands="" 
 
 [target.'cfg(windows)'.dependencies]
 winapi-util = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,8 +30,7 @@ unicode-width = "0.1"
 ignore="0.4"
 crossbeam-channel = "0.4"
 walkdir="2.3"
-# todo use num_format instead
-thousands="" 
+thousands = "0.2"
 
 [target.'cfg(windows)'.dependencies]
 winapi-util = "0.1"
--- a/src/display.rs
+++ b/src/display.rs
@@ -14,6 +14,7 @@ use std::cmp::min;
 use std::fs;
 use std::iter::repeat;
 use std::path::Path;
+use thousands::Separable;
 
 static UNITS: [char; 4] = ['T', 'G', 'M', 'K'];
 static BLOCKS: [char; 5] = ['█', '▓', '▒', '░', ' '];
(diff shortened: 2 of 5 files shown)


src/display.rs
```diff
@@ -323,12 +336,26 @@ fn get_name_percent(
...
+            first_size_bar
+        }
     } else {
-        pretty_size
+        let pretty_size = format!("{:>5}", human_readable_number(node.size));
+        if is_biggest && display_data.colors_on {
+            format!("{}", Red.paint(pretty_size))
+        } else {
+            pretty_size
+        }
```
bootandy, line 358: code duplication here ^

Cargo.toml
```diff
@@ -30,6 +30,8 @@ unicode-width = "0.1"
 ignore="0.4"
 crossbeam-channel = "0.4"
 walkdir="2.3"
+# todo use num_format instead
```
bootandy, line 33: https://crates.io/crates/num-format 

numformat  has way more downloads than thousands and is more configurable, we should probably use that.

src/display.rs
```diff
@@ -143,12 +146,20 @@ pub fn draw_it(
...
     }
-    let terminal_width = (get_width_of_terminal() - 14) as usize;
+    let num_chars_needed_on_left_most = if by_filecount {
+        let max_size = root_node.children.iter().map(|n| n.size).fold(0, max);
+        max_size.separate_with_commas().chars().count()
+    } else {
+        5 // Under normal usage we need 5 chars to display the size of a directory
+    };
+
+    let terminal_width = get_width_of_terminal() as usize - 9 - num_chars_needed_on_left_most;
```
lespea, line 162: Shouldn't this be 9 + num_chars_needed_on_left_most?

Cargo.toml
lespea: Yeah that'd probably be fine.  I actually was going to use that initially but figured I would go with a simpler crate.  They basically do the same thing for our usecase

src/display.rs
lespea: Why don't you have the if-block return the string to format, then do the formatting outside the if block?  Would remove the duplication

src/display.rs
bootandy: I think it depends how you read it
```(terminal_width - 9 ) - nums_chars_need```
vs
```terminal_width - (9 + nums_chars_need)```

lespea: Oh, right

Cargo.toml
bootandy: Frustatingly it looks like num-format doesn't build on a default windows Travis-ci box due to a weird clang dependency. 

I wanted num_format so 10,000 would be written as 10.000 in countries that use '.' instead of ',' but if it makes the CI this much bother I'll probably go back to the thousands crate.

nu8:
https://github.com/bootandy/dust/issues/103

PR #110, opened 2020-08-24
Started working on #103

skibaa:
Probably there is more work to do for this PR to be ready to merge, but decided to share my work early so I can get some feedback is the direction good enough.


skibaa:
Oh I see there is already progress on #104 and #109, so this PR is not relevant anymore.

PR #120, opened 2020-10-05
Be less prescriptive for the ansi_term dep

sylvestre:
It makes life of packager harder.
cargo test works with ansi_term 0.12.1


bootandy:
things like this make me nervous, but I'll trust you.

PR #121, opened 2020-10-05
Cleanup

baile320:
Hi, I'm practicing Rust and just wondering if you'd be interested in a PR with some cleanup changes? I was just reading through the code and noticed some things that might make things more readable or idiomatic, so I thought I would throw a PR out there.

There might be more stuff I can find (or go further w/ the few changes I made already - probably there are better ways to do some of them) but I wanted to see if you're interested in accepting this type of PR before I investigate further. Thanks :)

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -232,7 +232,8 @@ fn display_node(node: Node, draw_data: &DrawData, is_biggest: bool, is_last: boo
         percent_bar: bar_text,
         display_data: draw_data.display_data,
     };
-    let num_siblings = node.children.len() as u64;
+
+    let num_siblings = node.num_siblings();
 
     for (count, c) in get_children_from_node(node, draw_data.display_data.is_reversed).enumerate() {
         let is_biggest = dd.display_data.is_biggest(count, num_siblings);
@@ -398,7 +399,7 @@ mod tests {
             by_filecount: false,
             num_chars_needed_on_left_most: 5,
             base_size: 1,
-            longest_string_length,
+            longest_string_length: longest_string_length,
             ls_colors: LsColors::from_env().unwrap_or_default(),
         }
     }
--- a/src/display.rs
+++ b/src/display.rs
@@ -32,32 +32,16 @@ pub struct DisplayData {
 }
 
 impl DisplayData {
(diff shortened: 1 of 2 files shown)


src/utils/mod.rs
```diff
@@ -46,6 +46,22 @@ impl PartialEq for Node {
...
+    pub fn num_siblings(&self) -> u64 {
+        self.children.len() as u64
+    }
+
+    pub fn get_children_from_node(&self, is_reversed: bool) -> impl Iterator<Item = Node> {
+        let children = self.children.clone();
+        if is_reversed {
+            let children: Vec<Node> = children.into_iter().rev().collect();
+            return children.into_iter();
+        }
```
bootandy, line 59: if you add an 'else' on this line then we don't need the return on line 58. @baile320

src/utils/mod.rs
```diff
@@ -46,6 +46,22 @@ impl PartialEq for Node {
     }
 }
 
+impl Node {
+    pub fn num_siblings(&self) -> u64 {
+        self.children.len() as u64
+    }
+
+    pub fn get_children_from_node(&self, is_reversed: bool) -> impl Iterator<Item = Node> {
+        let children = self.children.clone();
```
bootandy, line 55: Do we need the clone of children here? If you think we do - then that's probably what the map was doing in the original method. 

 @baile320

src/display.rs
```diff
@@ -32,32 +32,16 @@ pub struct DisplayData {
...
-            } else if has_children {
-                "├─┬"
-            } else {
-                "├──"
-            }
+        match (self.is_reversed, was_i_last, has_children) {
+            (true, true, true) => "┌─┴",
+            (true, true, false) => "┌──",
+            (true, false, true) => "├─┴",
+            (true, _, _) => "├──",
```
bootandy, line 40: @baile320 lets make this line (true, true true) so it looks like the others.

bootandy:
Hi thanks for this,

Yes On the whole I think this looks sensible. 

Creating a node class makes sense. 

And while I had mixed feelings about the display.rs file initially, I do think your change makes things clearer. 

If you address the comments I'll merge it .

PR #130, opened 2021-01-13
Print correct message when paths are not found

### LINKED ISSUE #90: Give a sensible error on nonexistent directories
```
$ ls foo
ls: foo: No such file or directory
$ dust foo
Did not have permissions for all directories
$ 
```
This error message is wrong. I'd expect "no such file or directory" instead like ls.

alexmaco:
Fixes #90


bootandy:
This looks like a good idea, I'll re-read it when I have more time.

Issue #126, opened 2020-11-13
Support --width flag for non-terminal output


Nemo157:
I want to show `dust` output inside some environments where it is not possible to read the terminal width. The simplest example is if piping the output into a pager like `less`:

![temp](https://user-images.githubusercontent.com/81079/99066952-22983300-25aa-11eb-8486-ba41a1c02f8a.png)

These environments normally have variables available that can be passed into the command, so being able to specify something like `--width=$COLUMNS` would be useful.

bootandy:
yep I'm happy to add this flag.

Issue #128, opened 2020-12-31
when I am using tmux for splitting the terminal, dust output is weird


monacv:
could you please suggest a fix for $ dust . to have results proportial to size of splitted window in tmux?

![Screenshot from 2020-12-31 15-03-13](https://user-images.githubusercontent.com/76495162/103424331-86d27e80-4b79-11eb-834f-8be78d5fec9f.png)

bootandy:
I didn't know about this, I'll play with tmux and see if I can figure out what is going on

bootandy:
I tried splitting the screen on tmux and dust works correctly on my box.

I'm guessing that dust isn't able to get your screen width.

What happens if you run a simple rust program like this to get the width of your screen? Is the number incorrect?

```

use terminal_size::{terminal_size, Height, Width};

fn get_width_of_terminal()  {
    if let Some((Width(w), Height(_h))) = terminal_size() {
      println!("{:?}", w);
    }
 }
    

```

PR #131, opened 2021-01-16
Allow users to specify width of terminal

bootandy:


bootandy:
Fixes:
https://github.com/bootandy/dust/issues/126

May help with:
https://github.com/bootandy/dust/issues/128

PR #133, opened 2021-01-18
Compile and "function" under wasi

jcaesar:
Sorry for the noise: This is a PR I expect you not to merge, but to close.

Reasons being
* It requires some unpublished ~~/ unmerged~~ changes to lscolors ~~and terminal-size~~
* Getting `util` to function in an environment without threads, without slowing down all other "proper" platforms, requires somewhat nasty code changes
* Some wasi implementations expose incorrect / 0 values for dev/inode, so the results may be incorrect
* wasi doesn't expose a block size, so the results are definitely incorrect / always in apparent-size mode

The reasons I'm still making this PR are that
1. To show that somebody was "bored" enough to do it. ;)
2. To avoid anybody else getting the same idea, checking the issues for wasm/wasi, and not finding this
3. In the off case that you're super-enthusiastic about wasm/wasi, you can tell me and I will reopen this once the dependency changes have been merged and published.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -154,6 +154,7 @@ dependencies = [
  "ignore",
  "lscolors",
  "num_cpus",
+ "rustc_version",
  "stfu8",
  "tempfile",
  "terminal_size",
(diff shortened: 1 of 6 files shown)


bootandy:
thanks for looking at this @jcaesar I'll do my best to remember should wasi come up again.

PR #146, opened 2021-05-30
Updated crossbeam-channel to 0.5 and ran Cargo update

thomastay:
Hi! I use dust a lot, and was looking into the code. Hope its alright if I introduce some code cleanups:
PR number (1) is a tiny PR to update the dep of crossbeam_channel to the latest of 0.5.1, as well as run Cargo update. 
https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-channel/CHANGELOG.md#version-051

Looking at the changelog, there are no incompatibilities for 0.5.
This reduces the number of duplicated dependencies in the Cargo lock.


bootandy:
thanks

PR #147, opened 2021-06-05
Beautify help messages

QuarticCat:
Before:

image

After:

image

If you add `AppSettings::ColoredHelp`, the output can be better:

image

However, currently this cannot work with your `--no-colors` flag. See 
https://github.com/clap-rs/clap/issues/1880 and https://github.com/clap-rs/clap/issues/1185.

Actually, this not only affects help messages, but also affects error output of clap. (BTW, your current error handling is not consistent with clap's built-in. If you want a more consistent error output, you can check [clap::Error](https://docs.rs/clap/2.33.3/clap/struct.Error.html). But that will also introduce [some problems](https://github.com/ducaale/xh/pull/151#issuecomment-855223285).)

One compromise is to use `NO_COLOR` environment variable to disable colors. If you like that, I can commit relative codes.


QuarticCat:
I would like to say that `--no-colors` is not a good design. Usually we prefer automatically determine outputting colors or not. And add a flag `--color=always|auto|never` to control that. Although this is also hard to implement for clap's built-in color output.

bootandy (approved):
That's cool I didn't know you could do that.

PR #155, opened 2021-06-28
Add -ignore option to usage info in README

gmolau:
Had to Google to find out that obeying .gitignore files is already supported, so might as well add it to the usage info. (Might also be good to just put the output of `dust --help` into the README.)


bootandy:
Ah sorry, I've just removed this functionality with the 0.6.0 release. 

Inorder to make things run a lot faster I stopped using the 'ignore' crate which has this functionality.

Issue #158, opened 2021-07-09
Size of symlinked files rather than size of symlink


wdecoster:
Hi,

Great tool! I found this on https://github.com/ibraheemdev/modern-unix and want to use it on my system.

I am not sure if this is related to #157, but I noticed that dust also counts the file size of the file to which a symlink is pointing, rather than the size of the symlink.
For a directory with symlinks to pretty large files on another filesystem `dust` reports 22T, while `du` just returns 46M.
If I am looking in my home directory for things that use disk space I am not so much concerned about things that actually live on the huge storage server.

Is there already a switch to turn this off? Or is this the intended behavior?

Thanks,
Wouter

bootandy:
Thanks, 

Are the links soft links or hard links? 

If they are soft links I think that is a bug, I did a large refactor recently and if you run dust on a directory tree with soft links pointing to outside that directory tree then it takes the size of the file instead of the size of the symlink.

wdecoster:
Those are soft links. So yes, the soft links point to another file system and thus outside the directory tree.

AckslD:
Is it possible to tell it to report the size of the symlink with some argument?

bootandy:
This is a bug. I will fix it.

bootandy:
fixed in v0.6.1

Issue #157, opened 2021-07-08
Re-introduce -x to limit by filesystem


bootandy:
This option was removed during a rewrite as it was tricky to keep with the new design.

However, following this comment I think people will want it back:
https://github.com/bootandy/dust/commit/1d062cf683ee573ff1e629e257da86d89d92a00d

bootandy:
fixed in v0.6.1

PR #160, opened 2021-07-15
Change size of softlinks to 0

bootandy:
Instead of the size of what they point at


bootandy:
Fixes:
https://github.com/bootandy/dust/issues/158
https://github.com/bootandy/dust/issues/157

PR #172, opened 2021-08-20
Feature:  adding file types filter

bootandy:
-t = show summary of types
-y txt = show disk spaced taken only of .txt files
-y txt -f = show number of .txt files


bootandy:
superseded by a branch that allows regex:
https://github.com/bootandy/dust/pull/174

Issue #171, opened 2021-08-09
Filter by file type/extension


polarathene:
I often find support for exclude/ignore, but these tools seem to lack the opposite of allowing to only care about certain extensions (eg images `png|gif|jpg|jpeg|webp` or archives `zip|tar|rar|7zip|zstd`). Such functionality would be quite handy for me to get a better perspective of content breakdown for a server with user uploaded content.

Identifying how much disk usage a particular image type is using, or what the largest N images are. CLI tools like this one all seem to have the capability of doing such, but when the scanned content is not as well organized, or spread out among dirs where sibling/nested dirs can contain various other content or mixed types (eg if I only want to know about png disk usage), this is problematic without the ability to filter.

bootandy:
That's an interesting idea. I think it might be worth exploring. I'll look in to it.

We could group by extension type as well as a specific extension. 

https://www.computerhope.com/issues/ch001789.htm
https://www.online-convert.com/file-type

bootandy:
Here is an experimental branch where I've added -t for listing file_types and -y for a specific file type:

https://github.com/bootandy/dust/tree/by_type2

Is this what you were hoping for?

> 
> 
> $ target/release/dust       -y png                                                                                             [17:21:19]
>   64K   ┌── media            │                                                                   ██████████████████████████████████ │  33%
>   64K   │     ┌── media      │                                 ▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒██████████████████████████████████ │  33%
>   64K   │   ┌─┴ du-dust-0.6.1│                                 ▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒██████████████████████████████████ │  33%
>   64K   │   │ ┌── media      │                                 ▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒██████████████████████████████████ │  33%
>   64K   │   ├─┴ du-dust-0.6.2│                                 ▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒██████████████████████████████████ │  33%
[...]

PR #174, opened 2021-09-04
Feature: Adding file types filter & F flag changed

bootandy:
-t = Show summary of types

-e = Filter by regex
	allows you to specify a file type like -e "\.txt$"

Change behaviour of '-f' flag - it now counts only files. Before it
counted files & directories. This was needed for compatibility with
the new '-e' filter flag


bootandy:
For issue: https://github.com/bootandy/dust/issues/171

Issue #220, opened 2022-05-23
[WINDOWS]: The argument '--number-of-lines <number_of_lines>' cannot be used with '--depth <depth>'


yahiaboudah:
I want my depth to be 1 so I do: dust -d 1 -r -n Get-File-Count

Get-File-Count is a special powershell function I wrote to get the number of files and dirs in the current directory.

So yeah, it's not working

bootandy:
You can't use both -n and -d. Use one or the other not both.

yahiaboudah:
Can I specify how many directories and files are displayed, say when depth = 5? because it only shows a limited number of dirs, instead of the full length of all available directories and files.

bootandy:
Not currently, I was a bit unsure how it would work if a user provided both depth=5 and number=10. Thinking about it now I don't see why it isn't possible. I'll reopen this issue and take another look when I get time.

bootandy:
also by default using '-d' will show all the directories/file so using '-d 1' means you won't need a -n Get-File-Count

PR #224, opened 2022-07-01
Fix: Allow -n to be used with -d

bootandy:
Allow -n for number_of_lines to be used with -d 'max depth'

Remove depth specific functions, the job is now handled by the mainline

Add depth as a field onto the Node object.

https://github.com/bootandy/dust/issues/220

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -69,19 +69,19 @@ pub fn get_all_file_types(top_level_nodes: Vec<Node>, n: usize) -> Option<Displa
 
 fn add_children<'a>(
     using_a_filter: bool,
-    line: &'a Node,
+    file_or_folder: &'a Node,
     depth: usize,
     mut heap: BinaryHeap<&'a Node>,
 ) -> BinaryHeap<&'a Node> {
-    if depth > line.depth {
+    if depth > file_or_folder.depth {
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -236,26 +229,36 @@ fn main() {
...
     };
 
-    let depth = options.value_of("depth").and_then(|depth| {
-        depth
-            .parse::<usize>()
-            .map(|v| v + 1)
-            .map_err(|_| eprintln!("Ignoring bad value for depth"))
-            .ok()
-    });
+    let number_of_lines = match options.value_of("number_of_lines") {
```
bootandy, line 252: try using value_of_t again and accepting a default value.

src/filter.rs
```diff
@@ -78,16 +70,19 @@ pub fn get_all_file_types(top_level_nodes: Vec<Node>, n: usize) -> Option<Displa
 fn add_children<'a>(
     using_a_filter: bool,
     line: &'a Node,
```
bootandy, line 72: bad variable name

src/main.rs
bootandy: didn't work as we build the default dynamically based on value of depth.

PR #232, opened 2022-07-31
Lots of code cleanup

Lucretiel:
- Try to use iterator adapters and collect in various places, where possible. This especially benefits `draw_it`.
- Try to use `.map` and other similar methods on `Option`s and `Result`s, where possible
- Replaced nearly all `clone`s with reference-based equivalents
- Summarizing nodes by file extension is now much more efficient (significantly fewer allocations)
- `PartialOrd` and `PartialEq` implementations on `Node` and `DisplayNode` now agree with each other (it's no longer possible to construct a `Node` for which `node1 != node2 && !(node1 < node2) && !(node1 > node2)`).
- Replace `#[cfg(...)]` function definitions with simpler if `cfg!(...)` equivalents
    - The optimized code is the same and this reduces maintenance burden of keeping signatures synchronized
- Simplify CLI `Values` handling by taking advantage of Values::default
- Various spelling corrections in comments
- Replace (`sort`, `reverse`) with `sort_by`

I can split this into several PRs if you'd like, but each change was so small that it was easier to put them all together.

### DIFF
--- a/src/dir_walker.rs
+++ b/src/dir_walker.rs
@@ -34,14 +34,11 @@ pub fn walk_it(dirs: HashSet<PathBuf>, walk_data: WalkData) -> (Vec<Node>, bool)
     let top_level_nodes: Vec<_> = dirs
         .into_iter()
         .filter_map(|d| {
-            let n = walk(d, &permissions_flag, &walk_data, 0);
-            match n {
-                Some(n) => {
-                    let mut inodes: HashSet<(u64, u64)> = HashSet::new();
-                    clean_inodes(n, &mut inodes, walk_data.use_apparent_size)
-                }
-                None => None,
-            }
+            clean_inodes(
+                walk(d, &permissions_flag, &walk_data, 0)?,
+                &mut HashSet::new(),
+                walk_data.use_apparent_size,
+            )
         })
         .collect();
     (top_level_nodes, permissions_flag.into_inner())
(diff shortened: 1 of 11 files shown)


src/dir_walker.rs
```diff
@@ -34,14 +34,11 @@ pub fn walk_it(dirs: HashSet<PathBuf>, walk_data: WalkData) -> (Vec<Node>, bool)
...
-                    let mut inodes: HashSet<(u64, u64)> = HashSet::new();
-                    clean_inodes(n, &mut inodes, walk_data.use_apparent_size)
-                }
-                None => None,
-            }
+            clean_inodes(
+                walk(d, &permissions_flag, &walk_data, 0)?,
+                &mut HashSet::new(),
+                walk_data.use_apparent_size,
+            )
```
bootandy, line 41: nice!

bootandy:
Overall I think this is a good change. If you want to respond to my comments and fix it to build on Linux I'll be happy to accept. 

Thanks!

Lucretiel:
Done, that should fix everything, thanks for the review!

bootandy:
The mac builds failing is a new unrelated issue that seems to be caused by the temporary files being created in directories with very long names, I'm adding the '-w' param to the test to fix this.

bootandy:
I'm beginning to think we should break this PR up into a few different PRs

Lucretiel:
Fixed windows issue & mac test conflict

Issue #216, opened 2022-03-29
Select a file as baseline rather than total


Brooooooklyn:
For example, I want to compare the compress ratio between different algorithms

<img width="1490" alt="image" src="https://user-images.githubusercontent.com/3468483/160574543-1a733228-5bb5-49a8-b8e3-9e9fbec52550.png">

kidonng:
I would like to see this as well, similar to `du` without `--total`.

bootandy:
Added `--skip-total` flag which will not display the total and will instead base the bar graph on the largest file.

PR #239, opened 2022-08-18
Feature: Add skip-total flag

bootandy:
Flag to not include the last line containing totals of the output tree

resolves: https://github.com/bootandy/dust/issues/216


bootandy:
https://github.com/bootandy/dust/issues/216

PR #248, opened 2022-08-22
Support for completions

Jedsek:
Again :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,10 +84,18 @@ dependencies = [
  "bitflags",
  "clap_lex",
  "indexmap",
- "once_cell",
  "strsim",
  "termcolor",
  "textwrap",
+]
+
+[[package]]
+name = "clap_complete"
+version = "3.2.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e4179da71abd56c26b54dd0c248cc081c1f43b0a1a7e8448e28e57a29baa993d"
+dependencies = [
+ "clap",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


Jedsek:
I help you reduce clap's feature, because `crate_version!` is just a simple macro:

```rust
#[cfg(feature = "cargo")]
#[macro_export]
macro_rules! crate_version {
    () => {
        env!("CARGO_PKG_VERSION")
    };
}
```

Just replace it by using `env!`

Jedsek:
And now, no clippy warnings :)

Jedsek:
Also, I changed the `edition` form `2018` to `2021` in `Cargo.toml` and it could be compiled successfully.

bootandy:
Looks like the old 32 bit builds are broken.  (i686)  ( Could this be the change from 2018-> 2021 ? )

Jedsek:
Oh, maybe a place that I changed.

bootandy:
I don't think its the edition change as I opened a separate PR for that: https://github.com/bootandy/dust/pull/249

Jedsek:
I fix the max depth.
(I just thought about 64 bit before)

Jedsek:
Because a proble of clap, we could't set the default value of depth (When use a function which return value is a `Command`, because lifetime, see https://github.com/clap-rs/clap/issues/2150)

So I changed the `build_cli`'s signature to accept a new argument: `max_depth` to avold this problem. Otherwise, I think I have to add a dependencies `const_format` to achieve the same purpose.

Jedsek:
@bootandy All checks have passed, this pr could be merged

Issue #256, opened 2022-08-24
panicked at 'called `Result::unwrap()`


AminurAlam:
```
$ export RUST_BACKTRACE=full

$ dust
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ThreadPoolBuildError { kind: IOError(Os { code: 11, kind: WouldBlock, message: "Try again" }) }', src/main.rs:161:10
stack backtrace:
   0: 0xe7c35e34 - <unknown>
   1: 0xe7bd64be - <unknown>
   2: 0xe7c1f300 - <unknown>
   3: 0xe7c36836 - <unknown>
   4: 0xe7c3709c - <unknown>
   5: 0xe7c36e2a - <unknown>
   6: 0xe7c36df6 - <unknown>
   7: 0xe7c36dcc - <unknown>
   8: 0xe7bd4e58 - <unknown>
   9: 0xe7bd9ea6 - <unknown>
  10: 0xe7ba09cc - <unknown>
  11: 0xe7b87540 - <unknown>
  12: 0xe7ba13b6 - <unknown>
  13: 0xe78be0b6 - <unknown>

$ dust -V
Dust 0.8.2
```
this started happening after the update
installed dust from my package manager on termux

bootandy:
:-(.

Thanks for the info. 

From the line you've provided it seems Rayon is having trouble getting the threads. 

I'm doing a naked unwrap on that line so I'll change it to a log to stderr and handle the error.

bootandy:
https://github.com/rayon-rs/rayon/issues/694

bootandy:
Does this always happen or only sometimes? Is this machine a single core or is there anything odd about the machine you are running on? Thanks, @AminurAlam

AminurAlam:
this happens everytime, and i'm running on android 8.1 with 8 cores

bootandy:
@AminurAlam does the code here fix it? https://github.com/bootandy/dust/pull/259 ?

WesleyAC:
I'm seeing this every time on a single-cpu DigitalOcean droplet. If you can send me a binary version of #259, I'd be happy to try it out, @bootandy.

AminurAlam:
> @AminurAlam does the code here fix it? #259 ?

nope still broken
![20220829_011939](https://user-images.githubusercontent.com/64137875/187091946-ae3bbaf5-0dbf-4776-a90f-b57044997538.jpg)

bootandy:
[...]

PR #258, opened 2022-08-25
Fix error on thread creation2

bootandy:
Should Fix: https://github.com/bootandy/dust/issues/256


AminurAlam:
doesn't seem to fix the issue
```
$ git clone --depth 1 --branch 'fix_error_on_thread_creation2' 'https://github.com/bootandy/dust/'
$ cd dust
$ cargo run
```

first one is from repository
second one is compiled from source
third one is compiled from this pr

bootandy:
Ok, thanks for the information let me try again.

PR #266, opened 2022-09-25
Allow reading `inputs` from stdin

cstyles:
Sometimes it's nice to run `dust` on a list of files/directories generated programmatically with something like `fd` or read from a file. This PR allows users to do just that:

```shell
$ fd ... | dust
$ dust < list_of_directories.txt
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -213,6 +213,7 @@ version = "0.8.3"
 dependencies = [
  "ansi_term",
  "assert_cmd",
+ "atty",
  "clap",
  "clap_complete",
  "config-file",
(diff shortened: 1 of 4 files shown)


cstyles:
One note: `dust` doesn't like it when invoked like this:

```shell
$ echo | dust
Did not have permissions for all directories
```

However, it already behaves similarly when `inputs` is explicitly set to a list with a single, empty entry so it's not exactly a regression:

```shell
$ dust ''
Did not have permissions for all directories
```

bootandy:
$ echo | dust
Did not have permissions for all directories

I feel like that's the correct behaviour. You are passing the empty string into dust and dust is running on the empty string. It can not find a file or folder with the name emptystring so it prints that error message.

Issue #262, opened 2022-09-04
Add flag to only display directories


jrwrigh:
The description of the command inclines that only directories will be displayed, but on a machine I'm using it with it displays files as well:

```
3.9T   ├─┴ jrwrigh                                    │                                                 █████ │   8%
  0B   │ ┌── 2085581.output                           │                                             ░░░░░░░░█ │   0%
  0B   │ ├── 2085623.output                           │                                             ░░░░░░░░█ │   0%
  0B   │ ├── 2085643.output                           │                                             ░░░░░░░░█ │   0%
  0B   │ ├── 2085738.output                           │                                             ░░░░░░░░█ │   0%
  0B   │ ├── 2085756.output                           │                                             ░░░░░░░░█ │   0%
  0B   │ ├── 2085763.output                           │                                             ░░░░░░░░█ │   0%
  0B   │ ├── 2085764.output                           │                                             ░░░░░░░░█ │   0%
  0B   │ ├── 2085773.output                           │                                             ░░░░░░░░█ │   0%
  0B   │ ├── 2085793.error                            │                                             ░░░░░░░░█ │   0%
  0B   │ ├── 2085793.output                           │                                             ░░░░░░░░█ │   0%
4.0K   │ ├── 2085581.cobaltlog                        │                                             ░░░░░░░░█ │   0%
4.0K   │ ├── 2085581.error                            │                                             ░░░░░░░░█ │   0%
```

I'd like to not display any files, only directories and sub directories. For my particular use-case, I can get something close using `--min-size 1G` since I'm looking for large directories, but it's still catching large tar balls.

bootandy:
There is no flag to do this currently, it should be simple to add one. Consider this added to the todo list.

PR #268, opened 2022-09-30
FEATURE: support only directories will be displayed. Flag -D

hsnks100:
resolves: #262


src/filter.rs
```diff
@@ -41,15 +42,27 @@ pub fn get_biggest(
...
     file_or_folder: &'a Node,
     depth: usize,
     mut heap: BinaryHeap<&'a Node>,
 ) -> BinaryHeap<&'a Node> {
     if depth > file_or_folder.depth {
-        heap.extend(file_or_folder.children.iter().filter(|c| match min_size {
-            Some(ms) => c.size > ms as u64,
-            None => !using_a_filter || c.name.is_file() || c.size > 0,
-        }))
+        // heap.extend(file_or_folder.children.iter().filter(|c| c.name.is_dir()))
```
bootandy, line 51: delete this line and run cargo format on this file and we should be good to go :)

bootandy (changes requested):

src/filter.rs
hsnks100: I cleared code and cargo fmt.

I did rebase -i ... to clean up commit message.

PR #275, opened 2022-11-09
Implemented a progress indicator

NovaliX-Dev:
When dust is performing on a large set of file, it doesn't print anything to the terminal, which can confuse the user.

That's why i've implemented a small indicator. For the moment it doesn't show up if the process last less than 2 seconds (the timing is started at the indexing, because it's the operations that takes most of the time). If showed, it prints 3 things : 
- A rolling thing, to indicate that the program doesn't freeze
- The operation
- Some more infos if needed

It can show two operations : Indexing, and Preparing. When in Indexing, it shows the number of file indexed and the total size of them (the size is based on node's information). Preparing state is set after Indexing, when the nodes are cleanup and filtered before computing the output.

When the indicator is stopped, it returns the output, just in case you want to use them.

I'm pretty sure there are things that are not done in the best way, and i'll change them if requested, but for now i just wanted your opinion on this small feature.

I've also added a modification to your `.gitignore` file. You can reject it as well.


src/progress.rs
```diff
@@ -0,0 +1,308 @@
...
+    fn new(c: &PConfig) -> Self {
+        Self {
+            use_iso: c.use_iso,
+            ..Default::default()
+        }
+    }
+
+    fn format_size(&self) -> String {
+        let inner = self.inner.get();
+        let number_len = (inner as f32).log10().floor() as u32;
```
bootandy, line 119: could probably use  human_readable_number in display file.

src/dir_walker.rs
```diff
@@ -28,15 +33,21 @@ pub struct WalkData<'a> {
     pub ignore_hidden: bool,
 }
 
-pub fn walk_it(dirs: HashSet<PathBuf>, walk_data: WalkData) -> (Vec<Node>, bool) {
+pub fn walk_it(
+    dirs: HashSet<PathBuf>,
+    walk_data: WalkData,
+    info_data: Arc<PAtomicInfo>,
+    info_conf: Arc<PConfig>,
```
bootandy, line 40: I think the info_* variables should be added to WalkData struct (and probably made optional as I'd like a flag to turn off the in-progress message [although it should default to ON] )

bootandy (commented):
Overall I like it. 

I'd like the behaviour to sit behind a flag so someone can turn it off if required. But on the whole I think it's good. 

I've been offline for a while, but if you want to rebase I'd be happy to merge it in.

src/progress.rs
NovaliX-Dev: I've implemented this, but I think it's better if there is a space between the suffix and the text. What do you think about this ?

NovaliX-Dev:
I've finished modifying my PR.

src/progress.rs
bootandy: I think I prefer it without the space. That's what 'du' does. Plus for the tree view I need all the horizontal screen real estate I can get.

bootandy:
Looks pretty good. 

try a:
 cargo fmt & cargo clippy

NovaliX-Dev:
Cargo clippy and fmt done.

PR #280, opened 2022-12-12
Upgrade dependencies

marcograss:
upgrade all dependencies except clap (I saw in a comment you were maybe already working on switching to clap4)

it's pretty straightforward, only sysinfo required some code changes because they return the amount of avail memory in bytes now, not in kB.


bootandy:
Thanks I'll accept this but another PR updated the sysinfo dependency and now it conflicts. If you want to pull and rebase I'll accept.  

The style checks are because of a new version of rust/clippy - I'll fix that myself now.

Issue #269, opened 2022-10-17
Feature request: only list files and sort them by size


certik:
First of all, thank you for this program, it does exactly what I was looking for. In fact, I knew Rust will have this, so I just searched "rust cli better du", and it came up. :)

Here is what would be nice to have as a compiler option. Instead of showing this:
```
$ dust  
4.0K   ┌── cpptranslate.cpp   │                                         █ │   1%
4.0K   ├── example_input.txt  │                                         █ │   1%
4.0K   ├── parse.cpp          │                                         █ │   1%
4.0K   ├── parse2.cpp         │                                         █ │   1%
4.0K   ├── parse3.cpp         │                                         █ │   1%
8.0K   ├── dwarf_convert.py   │                                         █ │   2%
 12K   ├── CMakeLists.txt     │                                         █ │   2%
4.0K   │   ┌── whereami.h     │                                    ░░░░▒█ │   1%
 16K   │   ├── whereami.cpp   │                                    ░░░░██ │   3%
 20K   │ ┌─┴ whereami         │                                    ░░░░██ │   4%
8.0K   │ │ ┌── terminal_base.h│                                    ░░▒▒▒█ │   2%
 12K   │ │ ├── prompt0.h      │                                    ░░▒▒▒█ │   2%
 24K   │ │ ├── terminal.h     │                                    ░░▒▒██ │   5%
 44K   │ ├─┴ cpp-terminal     │                                    ░░████ │   9%
 64K   ├─┴ tpl                │                                    ██████ │  12%
 68K   ├── lfortran.cpp       │                                    ██████ │  13%
344K   ├── CLI11.hpp          │             █████████████████████████████ │  67%
516K ┌─┴ .                    │██████████████████████████████████████████ │ 100%
```
Show this:
```
4.0K   ┌── cpptranslate.cpp   │                                         █ │   1%
4.0K   ├── example_input.txt  │                                         █ │   1%
4.0K   ├── parse.cpp          │                                         █ │   1%
4.0K   ├── parse2.cpp         │                                         █ │   1%
4.0K   ├── parse3.cpp         │                                         █ │   1%
4.0K   │   ┌── whereami.h     │                                    ░░░░▒█ │   1%
8.0K   ├── dwarf_convert.py   │                                         █ │   2%
8.0K   │ │ ┌── terminal_base.h│                                    ░░▒▒▒█ │   2%
 12K   ├── CMakeLists.txt     │                                         █ │   2%
[...]

PR #306, opened 2023-01-28
feature: --files-only -F

bootandy:
Add feature to only show files skipping directories.

https://github.com/bootandy/dust/issues/269


bootandy:
https://github.com/bootandy/dust/issues/269

PR #310, opened 2023-02-13
Fix: some panics are occuring when creating rayon

bootandy:
Wrap and ignore these panics.

I can't reproduce this problem but I'm curious to know if this fixes it


kenden:
This should work (I didn't recompile) and is an improvement!
Still, getting the cryptic message:
`Problem initializing rayon, try: export RAYON_NUM_THREADS=1")`
on the first run of `dust` could be further improved.

PR #315, opened 2023-02-28
feat: default option for depth from config file

### LINKED ISSUE #314: Default depth value from config
The only thing blocks me to alias `dust` to `du` is the `depth` value. I don't want to specify `depth` every single time, and if I forget to add it can be painful if the directory structure is very deep. Most of the time I need `--depth=1`, but not all the time.

I can't set alias to `du="dust --depth=1"` because if I want to use a different value, it will cry about it: "The argument '--depth <depth>' was provided more than once".

The easiest solution would be to add a default value in the config (`depth`) and it would work the same way as other options:
1. If `--depth` is defined, use that value.
2. If `depth` is defined in config, use that value.
3. Otherwise use `usize::MAX` (that's the default now).

Is there a way to achieve it and I just couldn't figure it out or it's a missing feature. If it's a missing feature, I have a branch with it locally so I can file a PR anytime.

yitsushi:
- If `--depth` flag is not defined (or it has an invalid value), a value from the config file will be used.
- If no `depth` entry in the config file (or there is no config file), the default `usize::MAX` will be used.

Added test cases:
- no config and no flag defined              -> `usize::MAX` should be used
- config defined, but flag is not defined    -> config value should be used
- config is not defined, but flag is defined -> flag value should be used
- both config and flag is defined            -> flag value should be used

Additional changes:

- Fixed some clippy issues.
- Added comments to the example `config.toml` file. (copy from flag description)

Closes #314

### DIFF
--- a/config/config.toml
+++ b/config/config.toml
@@ -3,11 +3,29 @@
 #   ~/.config/dust/config.toml
 #   ~/.dust.toml
 
+# Print tree upside down (biggest highest)
 reverse=true
+
+# Subdirectories will not have their path shortened
 display-full-paths=true
+
+# Use file length instead of blocks
 display-apparent-size=true
+
+# No colors will be printed
 no-colors=true
+
+# No percent bars or percentages will be displayed
 no-bars=true
+
+# No total row will be displayed
 skip-total=true
+
+# Do not display hidden files
 ignore-hidden=true
+
+# print sizes in powers of 1000 (e.g., 1.1G)
 iso=true
+
+# Depth to show
+depth=1
--- a/config/config.toml
+++ b/config/config.toml
@@ -26,6 +26,3 @@ ignore-hidden=true
 
 # print sizes in powers of 1000 (e.g., 1.1G)
 iso=true
-
-# Depth to show
-depth=1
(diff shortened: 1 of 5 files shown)


src/display.rs
```diff
@@ -428,7 +428,7 @@ mod tests {
             iso: false,
         };
         DisplayData {
-            initial: initial,
+            initial,
```
yitsushi, line 431:
```
warning: redundant field names in struct initialization
```

config/config.toml
```diff
@@ -3,11 +3,29 @@
...
 skip-total=true
+
+# Do not display hidden files
 ignore-hidden=true
+
+# print sizes in powers of 1000 (e.g., 1.1G)
 iso=true
+
+# Depth to show
+depth=1
```
bootandy, line 31: I might remove this from the default config later - but you will be able to add it back in to your local config if you want it.

yitsushi: Oh that's the default config? I thought it's just an example with a few options.

yitsushi: Removed it from this file.

bootandy: actually this was my bad it was examples - sorry.

Issue #321, opened 2023-04-03
Is it possible to set the unit of measurement to a specified value,like KB?


TYzzt:
When the size of storage is large, the accuracy loses the angle

TYzzt:
[https://github.com/bootandy/dust/pull/322](url)

bootandy:
I'm not particularly keen on adding this flag. What is the usecase for it? 

If you need to know the exact size of a file you can use du for that. 

dust is going to ignore several files anyway with its targetted summary so I'm not sure what value this adds.

TYzzt:
> 

Fast and relatively accurate estimation of storage usage in HPC distributed storage scenarios

bootandy:
Hmm, maybe, let me sleep on it.

sergeevabc:
How do I get size in bytes? When I share files, I prefer to indicate the size in bytes so that people don't get confused by KB, KiB, etc. In addition, one can process the output in bytes later by other programs like awk and convert it into the desired units.

christosg88:
Please add a `--format` option, to specify the output's size format.

I want to parse the output from another script (in Python in my case) to merge it with other reports.
I prefer to call `dust` in a subprocess, and parse its output, because it's much faster than a native-Python solution.
But I don't want to turn my Python script which merges reports to Rust, because it would be too much work and it's already fast enough.

Any chance to implement that?

`dua` offers this option:

```
  -f, --format <FORMAT>
          The format with which to print byte counts

          [default: binary]
          [possible values: metric, binary, bytes, gb, gib, mb, mib]
```

bootandy:
Ok, clearly people want this. I'll look into it.

bootandy:
We could merge the --format flag with the --iso flag. - they seem to do similar things.

hydrasho:
I also need this feature. currently i'm forced to use du which is extremely slow and slows down an important script. i wanted to replace it with dust but it forces the -h (human readable). a --format or the inverse of a -h even if it's not precise would be great! thank !

bootandy:
https://github.com/bootandy/dust/pull/373

it's merged it'll be in the next release.

sergeevabc:
[...]

PR #322, opened 2023-04-10
#321: display the size of a file or directory in "kilobytes"

TYzzt:
added **-k** parameter to display the size of a file or directory in "kilobytes"


TYzzt:
hi, @bootandy, could you please help me review the code

bootandy:
your code looks good.

bootandy:
I think the code you have written is sensible. But I'm just not sure that we need to cover this usecase.

TYzzt:
> 

thank you~ The kilobyte formatting is very useful in scenarios where higher precision is required. This is also the requirement that my work encounters.

keenahn:
I would also like this, as well as bytes (aka the non human-readable version of `du`). I'll fork @TYzzt's branch :).

bootandy:
https://github.com/bootandy/dust/pull/372

Commits from this PR moved to the above PR

Issue #197, opened 2021-11-16
Aborted (core dumped)


taqtiqa-mark:
```bash
$ /home/user/.cargo/bin/dust --version
Dust 0.7.0
$ /home/user/.cargo/bin/dust                                   
                                                                               
thread '<unknown>' has overflowed its stack                         
fatal runtime error: stack overflow                                        
Aborted (core dumped)
```

bootandy:
ok, what were you running that on ?

taqtiqa-mark:
System: Ubuntu 18.04
Folder: ~/

bootandy:
Does it happens with old versions of dust ? 

if you 'cd' into one of the directories inside ~ do each of them work - if not which one fails?

Does du work ?

Do you have any suspicious directories that symlink to interesting places ? 

Aprox how big is '~' ?

BobIsOnFire:
Hi! I got exact same issue today. It reproduces kinda randomly and could be related to having a huge number of path components.

I was trying to `dust` my working directory, as usual. But this time, I had 500 directories one inside another I previously generated for tests (`mkdir -p $(for i in {1..500}; do echo -n "qwe/"; done)`)

OS: RHEL 8.4

```bash
17:23:11 ~$ ls repro
build  qwe  src  toolchain

17:23:15 ~$ dust -rbd0 repro

thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
17:23:16 ~$ dust -rbd0 repro

thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
17:23:17 ~$ dust -rbd0 repro

thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
17:23:17 ~$ dust -rbd0 repro

thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
17:23:17 ~$ dust -rbd0 repro
 9.3G └── repro

17:23:18 ~$ for i in $(seq 1 100); do dust -rbd0 repro/ ; done &>repro.log
17:23:44 ~$ grep -c 'stack overflow' repro.log
74
```

Backtrace (collected on 0.8.1):
```
#0  0x00007ffff708437f in raise () from /lib64/libc.so.6
#1  0x00007ffff706edb5 in abort () from /lib64/libc.so.6
[...]

PR #357, opened 2023-12-11
Stack size2

bootandy:


bootandy:
might help with: https://github.com/bootandy/dust/issues/197

Issue #359, opened 2023-12-28
zsh completion does not allow directories to be tab-completed


mdekstrand:
With the ZSH completion, when I attempt to tab-complete a directory:

```console
$ dust Docu<tab>
```

it does nothing. The expected behavior is for it to tab-complete the directory name.

bootandy:
$ echo $SHELL                                                                                                                                                                      /usr/bin/zsh

$ dust [TAB]

this works for me. 

Can you double check this or see if it works on bash ?

mdekstrand:
Just tried again, no dice.

I'm ZSH on MacOS, using Dust 0.8.6 and its completions installed by Homebrew. I also just tried installing the latest version of Dust (`brew install --HEAD dust`), and the problem persists.

bootandy:
code points:
https://github.com/bootandy/dust/tree/master/completions
https://github.com/bootandy/dust/blob/master/build.rs

commit:
https://github.com/bootandy/dust/commit/81d52e6e3a6aa5220eb19c72fbc4f9ad75fda7f2

I'm not sure what I can do here, I don't have a mac so I can't reproduce this - you are welcome to try and fix it yourself 

Perhaps try the command: rehash ?

wickles:
I had fixed this before in 0038cb24b4adbb85321e3c36fb7079fc898de44e but it looks like it got reverted in 2fa14ca19cd80ba2be106784e2626b64b199e61b.

bootandy:
thanks for your comment @wickles - I must have something that auto-updates and over-writes it. I'll investigate.

bootandy:
Which Z-shell version are you guys running ? 

`$ $SHELL --version  
5.8.1 (x86_64-ubuntu-linux-gnu)
`

I've no idea what to do here. I can't cause a failure on my system with my zshell. 
If I try and edit the completions files running 'cargo build' will reset them back to what they were before.

BuonOmo:
```
$ brew install --HEAD dust
$ dust --version
Dust 0.9.0
$ $SHELL --version
zsh 5.9 (x86_64-apple-darwin23.0)
$ dust ~/<tab><tab> # fails
$ which _dust
_dust () {
	typeset -A opt_args
	typeset -a _arguments_options
	local ret=1 
	if is-at-least 5.2
	then
		_arguments_options=(-s -S -C) 
	else
		_arguments_options=(-s -C) 
	fi
	local context curcontext="$curcontext" state line 
[...]

PR #368, opened 2024-02-11
Fix filename completions

bootandy:
I previously accidentally reverted a commit that @wickles made to fix the completion.

https://github.com/bootandy/dust/issues/359

### DIFF
--- a/completions/_dust
+++ b/completions/_dust
@@ -70,8 +70,7 @@ _dust() {
 '--help[Print help]' \
 '-V[Print version]' \
 '--version[Print version]' \
-'*::params:' \
-'*:inputs:_files' \
+'*::params:_files' \
 && ret=0
 }
 
--- a/completions/_dust
+++ b/completions/_dust
@@ -71,6 +71,7 @@ _dust() {
 '-V[Print version]' \
 '--version[Print version]' \
 '*::params:' \
+'*:inputs:_files' \
 && ret=0
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,9 +22,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.5"
+version = "0.6.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


wickles:
Maybe I'm wrong but I think there needs to be an adjustment in `build.rs` or `src/cli.rs`, or else `clap` will just revert this again with the next build.

completions/_dust
```diff
@@ -71,6 +71,7 @@ _dust() {
 '-V[Print version]' \
 '--version[Print version]' \
 '*::params:' \
+'*:inputs:_files' \
```
wickles, line 74: This will not work. Remove the new line and replace the existing line with
```
'*::params:_files' \
```

wickles (changes requested):

completions/_dust
bootandy: Happy to make this change. 

When I test tab completion it seems to work for me. 

How would you recommend that I test this? & How do you know this is the right thing to do ? 

Thanks

bootandy:
>Maybe I'm wrong but I think there needs to be an adjustment to the completion generation in build.rs and src/cli.rs, or else clap will just revert this again with the next build.

You are right, going to rethink this.

bootandy:
There's a lot of auto complete issues raised in clap-complete. https://github.com/clap-rs/clap/issues?page=2&q=is%3Aissue+is%3Aopen+completion

A (rather hacky) solution could be to add a comment saying 'do not commit this change' if clap rebuilds it. Although, ideally we want to fix the root problem which is either in clap-complete or my use of clap.

completions/_dust
wickles: Just edit the existing file in your fpath and then start a new zsh session. When I try with both `*:...` lines added:
```
❯ dust <tab>
_arguments:comparguments:327: doubled rest argument definition: *:inputs:_files
_arguments:comparguments:327: doubled rest argument definition: *:inputs:_files
❯ dust 
```
compared with the one line as in my suggestion:
```
❯ dust <tab>
-- params --
Desktop/  Documents/ ...
```

PR #370, opened 2024-02-20
feat(win): use size on disk for apparent size for OneDrive files

### LINKED ISSUE #366: Use Size on Disk on Windows
On Windows, Dust should use Size on Disk instead of Size, because OneDrive and other cloud storage providers can create files which do not take up space on the disk

pandaninjas:
Resolves #366 

Before: 
image
image
After: 
image
image

### DIFF
--- a/src/platform.rs
+++ b/src/platform.rs
@@ -90,7 +90,10 @@ pub fn get_metadata(d: &Path, use_apparent_size: bool) -> Option<(u64, Option<(u
         Ok(Handle::from_file(file))
     }
 
-    fn get_metadata_expensive(d: &Path, use_apparent_size: bool) -> Option<(u64, Option<(u64, u64)>)> {
+    fn get_metadata_expensive(
+        d: &Path,
+        use_apparent_size: bool,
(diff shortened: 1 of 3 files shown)


src/platform.rs
```diff
@@ -90,16 +90,24 @@ pub fn get_metadata(d: &Path, _use_apparent_size: bool) -> Option<(u64, Option<(
         Ok(Handle::from_file(file))
     }
 
-    fn get_metadata_expensive(d: &Path) -> Option<(u64, Option<(u64, u64)>)> {
+    fn get_metadata_expensive(d: &Path, _use_apparent_size: bool) -> Option<(u64, Option<(u64, u64)>)> {
```
bootandy, line 93: this doesn't need to be an '_' now as the parameter is now used

bootandy:
Also: can you run 'cargo fmt' please.

If it's a pain - don't worry I can accept this and clean it up.

bootandy:
Looks good. I'll accept in a couple of days to give you time to clean up first. Thanks.

bootandy:
Thanks!

PR #381, opened 2024-03-22
Better handling for color in terminal

Its-Just-Nans:
Hi

With this PR

```
dust | less
```

will work out of the box

It checks if the output is a terminal (if not, it could be a pipe, so we disable the colors)

This PR also implements the https://no-color.org/ proposal

If you still want the colors you can use

```bash
dust -C
# or
dust --force-colors
```

Thanks !


bootandy:
Yeah this looks sensible, run `cargo fmt` and I'll accept it - thanks.

Its-Just-Nans:
Forgot it, done :)

bootandy:
Getting there, but still missing a few

Its-Just-Nans:
Maybe this time ? (I hope)

 and I added two tests :)

bootandy:
I'll clean the windows one

Its-Just-Nans:
I missed it !

Thanks!

PR #389, opened 2024-04-25
Added json output function

fksms:
I added it because I needed the ability to output it as json.

`dust -j` outputs json to the current directory.

For your reference.


bootandy:
Ok, lets fix the format and go

fksms:
Thank you for confirmation.

What specifically should I fix?

bootandy:
Run `cargo fmt` locally - you can see the build failed above ^ , then push again.

bootandy:
and rebase

fksms:
Very helpful.

I've Done!

bootandy:
https://github.com/bootandy/dust/pull/396
Here, we print to stdout instead - I don't want my tool writing files to the disk. - if you want the old behaviour you can do 
`dust > node.json`

PR #393, opened 2024-04-28
Redo 'Fix zsh completion' with clap

eatsu:
Because I'm not familiar with Rust, when I was working on https://github.com/bootandy/dust/pull/390 I didn't realize that the completions were auto-generated via clap. I'm sorry. This redoes it with clap.

This improves not only the completions but also the --help and the man page. Also the --output-format flag will raise an error if the given value is invalid.


eatsu:
Since dust prints sizes with units `B`, `K`, `M`, etc instead of `B`, `KiB`, `MiB`, I've changed the primary values for `--output` to `k`, `m`, etc and made `kib`, `mib`, etc their (hidden) aliases. Please let me know if this should be reverted.

bootandy:
>Since dust prints sizes with units B, K, M, etc instead of B, KiB, MiB, I've changed the primary values for --output-format to k, m, etc and made kib, mib, etc their (hidden) aliases. Please let me know if this should be reverted!

Yeah, let's revert this. This was my fault and I'm working on fixing it in 
https://github.com/bootandy/dust/pull/392

bootandy:
Are you doing anything special to generate `completions/_dust` ? - When I run a `cargo build` it seems to remove the variable from that file, I wonder if I have something configured incorrectly  

```
-'-d+[Depth to show]:depth' \
+'-d+[Depth to show]: : ' \
```

eatsu:
> This was my fault and I'm working on fixing it in
https://github.com/bootandy/dust/pull/392

I'm sorry if my explanation was not clear enough, but I think this (one letter units are preferred everywhere) is unrelated with the bug dust currently has?

To be clear (from my viewpoint):

- 1024 by default -> good, no problem
- dust prints sizes using the one letter units (B, K, M, ...) -> good, no problem
- `-o k` should be based on 1024
- `-o kib` should be based on 1024
- `-o kb` should be based on 1000

> Are you doing anything special to generate completions/_dust ? - When I run a cargo build it seems to remove the variable from that file, I wonder if I have something configured incorrectly

No, I installed rust using rustup and did not configure anything in particular. I'll look into it.

bootandy:
>I'm sorry if my explanation was not clear enough, but I think this (one letter units are preferred everywhere) is unrelated with the bug dust currently has?

You are correct. I misread your early message.

PR #394, opened 2024-04-29
Allow options to be set in any position

eatsu:
Currently options following regular arguments are not interpreted as options. This fixes that.

Users can still treat values starting with a hyphen (`-`) as regular arguments by using `--`, e.g.:

    dust -d 2 -r ~/Documents -F -- --this-is-my-dir


bootandy:
Neat.

PR #396, opened 2024-05-02
refactor: cleanup -j / --output-json flag

bootandy:
Stop -j writing to a file, print to stdout instead.


bootandy:
TODO: Also if -j is supplied remove the spinner as that will pollute the stdout see the `--no-progress` option.

PR #397, opened 2024-05-12
eprint problematic folders

Its-Just-Nans:
This PR change the `Did not have permissions..` message to print problematic folders/files

Why this PR: I think it's useful to have the list of the problematic folders/files


bootandy:
I don't want to do this - the list would be too long and obscure the 'real' output.

Its-Just-Nans:
@bootandy what about adding a CLI flag to have the list?

I think the user would like to know where dust isn't working and check why it isn't !

bootandy:
Ok, I'll accept it if we add a CLI flag to control it.

Its-Just-Nans:
Thanks

I've added the flag

Issue #386, opened 2024-04-21
`--only-file` displays full paths even without `--full-paths`


eatsu:
First of all, thank you very much for this helpful tool.

I have an issue that `--only-file` displays full paths even without `--full-paths`, like this:

```console
$ dust --only-file /path/to/target-dir
...   ┌── /path/to/target-dir/foo         ...
...   ├── /path/to/target-dir/sub-dir/bar ...
... ┌─┴ /path/to/target-dir               ...
```

It would be nice if it could display the shortened, relative paths by default, like this:

```console
$ dust --only-file /path/to/target-dir
...   ┌── foo         ...
...   ├── sub-dir/bar ...
... ┌─┴ target-dir    ...
```

I'm using Dust 1.0.0 on Arch Linux.

bootandy:
That's a good point, I'll look into this.

PR #398, opened 2024-05-13
fix: issue 386 --only-file displays full paths even without --full-paths

wugeer:
#Background
--only-file displays full paths even without --full-paths #386
 If only-file is provided without the --full-pathss parameter, the output file path does not contain the full path

# Full Changelogs
- fix: --only-file displays full paths even without --full-paths

# Issue Reference
Fix https://github.com/bootandy/dust/issues/386 --only-file displays full paths even without --full-paths 

# Test Result
image

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -52,10 +52,9 @@ impl Config {
         Some(true) == self.ignore_hidden || options.get_flag("ignore_hidden")
     }
     pub fn get_full_paths(&self, options: &ArgMatches) -> bool {
-        // If we are only showing files, always show full paths
         Some(true) == self.display_full_paths
             || options.get_flag("display_full_paths")
-            || self.get_only_file(options)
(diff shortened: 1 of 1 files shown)


bootandy:
Hmm, I now remember why I did this originally. If we accept this then when we run with --only-file we get a list of files but we have no way of knowing which folder they are in. 

I'm unsure

wugeer:
When running with --only-file, we get a list of files, but we have no way of knowing which folder they are in. I think this is reasonable. This is the function of --full-paths to make the file output as an absolute path. :)

bootandy:
Ok, I'll accept, but I can't make any promises about not changing my mind in the future.

bootandy:
oops - we broke a test with this: fixing: https://github.com/bootandy/dust/pull/403

Issue #340, opened 2023-10-31
Filter by file age


MolbioUnige:
I am looking for a way to take in account only files that are older then a given value. Would that be possible to implement?

In my institue, people are making experiments that weight a few terabytes each. People are required to remove their data once their treatment is completed, but some tend to forget. So, I'd like to find a way to find who is sitting on old stuff and thus preventing other users to use the servers.

bootandy:
That is an interesting idea, 

It should be possible to implement. I'll leave this issue open

smf8:
Hi, As I've last checked with the master branch. This hasn't been implemented yet and I'm interested in helping out if that's OK.
I'm fairly new to Rust and so I apologize beforehand but I've got a couple of questions.

- Should this filtering happen in file walking stage? (like in `ignore_file` which takes some filtering (e.g. regex) into account)

- Or is it better to apply it when extracting `DisplayNode` from `Node`s?
 
I think the latter might be more beneficial since It can also address #385 by building the `heap` based on `age` instead of `size`.

bootandy:
That's a good question. 

In this case with a hard limit I would say: this filtering happen in file walking stage (probably).

However it might be fun to try it both ways. 

For https://github.com/bootandy/dust/issues/385 however it would probably come from when: extracting DisplayNode from Nodes. Even thought the tasks sound similar they might not be in code. 

I'm not sure of the above myself yet though, so I may well change my mind!

wugeer:
Hello, I'm fairly new to Rust and so I apologize beforehand but I've got a couple of questions. I am trying to deal with this problem, but I encountered a problem
- how to filter according to the time corresponding to the input, is it to filter the files of those days according to the sign of the input integer value like Linux find, or is it to input What about a format like --mtime '+2023-10-04 05:34:00'? Or what about epoch-based time like --mtime +1695276503? 

In other words, is it better to use relative time, absolute time, or epoch-based time as input?

codeswhite:
Hey, I had an idea,

Consider supporting suffix-parsing for the duration for example:
- `30s` for 30 seconds
- `5m` for 5 minutes
- `2h` for 2 hours
- and maybe even `2day4hours`

This way it will be much more intuitive and user friendly as well as powerful.

PR #399, opened 2024-05-17
feat: Added the ability to filter the corresponding files based on the access_time, modify time, and change time of the file for statistics.

wugeer:
# Background

Filter by file age https://github.com/bootandy/dust/issues/340
Add a new feature to dust to filter files based on their last update time, last access time, and last modification time.          

In order to achieve this requirement, I referred to the input of the linux find command, and ultimately planned to implement the dust filter by file age function based on the input style of find.

image

I have a little question here, why does dust output files or directories with a file size of 0 by default? The dust tool is used to quickly find those files and directories in the specified directory that take up too much space, and those empty files or directories will only affect the look and feel. Originally, I have implemented filtering out those files or directories with a size of 0, but when I run cargo test When running the test, I found that some samples reported errors, so I canceled this part of the code, but I still wanted to retain the function of filtering out those files or directories with a size of 0. I hope this function can be further explored. thanks

# Full Changelogs

- feat: Added the ability to filter the corresponding files based on the access_time, modify time, and change time of the file for statistics

# Issue Reference

fix:  Filter by file age https://github.com/bootandy/dust/issues/340

# Test Result
dust help command output
image

Test Data

image

cargo test run test results

image

Adding --mtime or --atime or --ctime results in actual execution of dust and linux find with similar parameters.

image

image

image

image

image


bootandy:
Apologies I've had a laptop failure. Don't want to risk looking at this on my work laptop for legal reasons. Will review this when I get a new machine.

completions/_dust
```diff
@@ -37,6 +37,12 @@ _dust() {
 '--output-format=[Changes output display size. si will print sizes in powers of 1000. b k m g t kb mb gb tb will print the whole tree in that size.]:FORMAT:(si b k m g t kb mb gb tb)' \
 '-S+[Specify memory to use as stack size - use if you see\: '\''fatal runtime error\: stack overflow'\'' (default low memory=1048576, high memory=1073741824)]:STACK_SIZE: ' \
 '--stack-size=[Specify memory to use as stack size - use if you see\: '\''fatal runtime error\: stack overflow'\'' (default low memory=1048576, high memory=1073741824)]:STACK_SIZE: ' \
+'-M+[File'\''s data was last modified less than, more than or exactly <mtime>*24 hours ago. When dust figures out how many 24-hour periods ago the file was last modified, any fractional part is ignored, so to match -mtime +1, a file has to have been modified at least two days ago.]: : ' \
```
codeswhite, line 40: Consider making the description more easily readable


```suggestion
'-M+[Sort by time modifed, include only files modified +/-<mtime> days before/after, days are rounded down, prefix \'\-\' for newer, \'\+\' for older. : : ' \
```

wugeer: In my opinion, Filter by time mofied is better than Sort by time modifed because this parameter is used for filtering, not sorting. right? :)

codeswhite: Sure, you can rephrase it as you wish, just pointed that its a bit too long :)

src/dir_walker.rs
```diff
@@ -167,6 +191,9 @@ fn walk(dir: PathBuf, walk_data: &WalkData, depth: usize) -> Option<Node> {
                                         data.is_file(),
                                         walk_data.by_filecount,
                                         depth,
+                                        filter_modified_time,
```
codeswhite, line 194: Consider changing `build_node` to accept a `walk_data` argument instead of individually 7 of `walk_data`'s fields

completions/_dust
wugeer: Indeed, that description is a bit too long. I rephrased it . I will be very happy if you can help me review the code again. :)

src/dir_walker.rs
wugeer: Indeed, you are right, I reorganized the code according to your ideas. I will be very happy if you can help me review the code again. :)

codeswhite (approved):
Looks good!
Should wait for @bootandy to review as well

I still suggest that after merging this we continue to work on a more robust implementation that will support all timeframes (like i explained in #340 )

wugeer:
> Looks good! Should wait for @bootandy to review as well
>
> I still suggest that after merging this we continue to work on a more robust implementation that will support all timeframes (like i explained in #340 ) All timeframes (like I explained in #340).

That sounds like a good idea. I like it! I'll try it. ：)

wugeer:
> Looks good! Should wait for @bootandy to review as well
>
> I still suggest that after merging this we continue to work on a more robust implementation that will support all timeframes (like i explained in #340 )

I have thought about it carefully these days. If I modify it according to your idea, I need to add a parameter to limit the upper and lower limits of the filtering time. Otherwise, there will be no equal scenarios. Although this method provides great flexibility, As for whether the tool itself needs such great flexibility, and whether adding this function will bring a lot of mental burden to the user, after all, everyone may have used it and are accustomed to the find command's -mtime, which is in units of days or - mmin is just filtering with two granularities, what do you think about it?> Looks good! Should wait for @bootandy to review as well
> 
> I still suggest that after merging this we continue to work on a more robust implementation that will support all timeframes (like i explained in #340 )

wugeer:
> Looks good! Should wait for @bootandy to review as well
>
> I still suggest that after merging this we continue to work on a more robust implementation that will support all timeframes (like i explained in #340 )

I've been thinking seriously about this over the past few days. If we modify the tool according to your approach, we'll need to add a parameter to set the upper and lower limits of the filter time. Otherwise, the scenario for equal values would no longer exist. While this method provides great flexibility, I'm concerned about whether such a high level of flexibility is necessary for this tool. Adding this feature might place a significant cognitive burden on the users. After all, people are probably accustomed to using the find command with the -mtime option, which filters by days, or the -mmin option, which filters by minutes. What are your thoughts on this? @bootandy

completions/_dust.ps1
```diff
@@ -43,6 +43,12 @@ Register-ArgumentCompleter -Native -CommandName 'dust' -ScriptBlock {
...
+            [CompletionResult]::new('-M', 'M ', [CompletionResultType]::ParameterName, '+/-n matches files modified more/less than n days ago , and n matches files modified exactly n days ago, days are rounded down.That is +n => (−∞, curr−(n+1)), n => [curr−(n+1), curr−n), and -n => (𝑐𝑢𝑟𝑟−𝑛, +∞)')
+            [CompletionResult]::new('--mtime', 'mtime', [CompletionResultType]::ParameterName, '+/-n matches files modified more/less than n days ago , and n matches files modified exactly n days ago, days are rounded down.That is +n => (−∞, curr−(n+1)), n => [curr−(n+1), curr−n), and -n => (𝑐𝑢𝑟𝑟−𝑛, +∞)')
+            [CompletionResult]::new('-A', 'A ', [CompletionResultType]::ParameterName, 'just like -mtime, but based on file access time')
+            [CompletionResult]::new('--atime', 'atime', [CompletionResultType]::ParameterName, 'just like -mtime, but based on file access time')
+            [CompletionResult]::new('-y', 'y', [CompletionResultType]::ParameterName, 'just like -mtime, but based on file change time')
+            [CompletionResult]::new('--ctime', 'ctime', [CompletionResultType]::ParameterName, 'just like -mtime, but based on file change time')
```
bootandy, line 51: I'm not sure if we need all these - I like the idea of a including a filter that works with modified time. But 'access time' - would someone actually use that? I have to launch 'stat' to even see it.

src/dir_walker.rs
```diff
@@ -160,13 +181,10 @@ fn walk(dir: PathBuf, walk_data: &WalkData, depth: usize) -> Option<Node> {
...
                                         entry.path(),
                                         vec![],
-                                        walk_data.filter_regex,
-                                        walk_data.invert_filter_regex,
-                                        walk_data.use_apparent_size,
                                         data.is_symlink(),
                                         data.is_file(),
-                                        walk_data.by_filecount,
                                         depth,
+                                        walk_data,
```
bootandy, line 187: sensible change, there were too many params.

bootandy (changes requested):
I like that you based the parameters for this work from find.

We seem to have a few cargo style problems. But overall I think this approach is correct.

completions/_dust.ps1
bootandy: Ah we can leave it in if you'd like.

wugeer:
> > I have a little question here, why does dust output files or directories with a file size of 0 by default? I have a little question, why does dust output files or directories with a file size of 0 by default?
>
> Just incase you are running on a very small folder. Some people seem to use dust as a sort of file-finder so I want to include them. Some people seem to use dust as a file finding tool, so I wanted to include them.

Yes, you are right! :)

wugeer:
> I like that you based the parameters for this work from find. I like that you based the parameters for this work from find.
>
> We seem to have a few cargo style problems. But overall I think this approach is correct. We seem to have a few cargo style problems. But overall, I think this approach is the right one.

I'm fairly new to Rust , so my coding style may not be very good. If there is anything that needs to be modified, I hope you will give me some advice. thx!

completions/_dust.ps1
wugeer: In my opinion, access time is an effective filtering tool when users want to determine whether a file has been accessed recently based on the access time to determine whether other people are using the file.

bootandy:
>I'm fairly new to Rust , so my coding style may not be very good. If there is anything that needs to be modified, I hope you will give me some advice. thx!

Your coding style looks fine to me. The build however gave style errors ^ see the red 'X's in the build. 

On your cmd line:

Run 'cargo fmt' - this neatens things automatically

Run 'cargo clippy' - this is a linter which will suggest fixes and warns of likely bugs, so you'll have to review each one. 

you can see the list of errors from the online build here: https://github.com/bootandy/dust/actions/runs/9461768640/job/26334984807?pr=399

wugeer:
> > I'm fairly new to Rust , so my coding style may not be very good. If there is anything that needs to be modified, I hope you will give me some advice. thx!
> 
> Your coding style looks fine to me. The build however gave style errors ^ see the red 'X's in the build.
> 
> On your cmd line:
> 
> Run 'cargo fmt' - this neatens things automatically
> 
> Run 'cargo clippy' - this is a linter which will suggest fixes and warns of likely bugs, so you'll have to review each one.
> 
> you can see the list of errors from the online build here: https://github.com/bootandy/dust/actions/runs/9461768640/job/26334984807?pr=399

Thank you for your guidance in helping me improve my code to meet the standards. I have resubmitted it, and all checks have passed. :)

bootandy:
> this method provides great flexibility, I'm concerned about whether such a high level of flexibility is necessary for this tool. Adding this feature might place a significant cognitive burden on the users. After all, people are probably accustomed to using the find command with the -mtime option, which filters by days, or the -mmin option, which filters by minutes. What are your thoughts on this? @bootandy

I don't have strong opinions on this. But lets stabilise -mtime an -mmin for now. If we want to do the more flexible one we can do that in a followup. It definitely shouldn't go in this PR.

bootandy:
Would you mind doing a rebase or merge - I've accepted a different PR and there are some conflicts ^ - thanks .

wugeer:
> > this method provides great flexibility, I'm concerned about whether such a high level of flexibility is necessary for this tool. Adding this feature might place a significant cognitive burden on the users. After all, people are probably accustomed to using the find command with the -mtime option, which filters by days, or the -mmin option, which filters by minutes. What are your thoughts on this? @bootandy
> 
> I don't have strong opinions on this. But lets stabilise -mtime an -mmin for now. If we want to do the more flexible one we can do that in a followup. It definitely shouldn't go in this PR.

Indeed, I agree with you. :)

wugeer:
> Would you mind doing a rebase or merge - I've accepted a different PR and there are some conflicts ^ - thanks .

Ok. I'm happy to do it! :)

src/config.rs
```diff
@@ -6,8 +7,11 @@ use std::io::IsTerminal;
 use std::path::Path;
 use std::path::PathBuf;
 
+use crate::dir_walker::Operater;
 use crate::display::get_number_format;
 
+pub static DAY_SECEONDS: i64 = 24 * 60 * 60;
```
bootandy, line 13: typo!

wugeer: Sorry, due to my carelessness, I did not correct this spelling error in time. I will resubmit a PR for processing.

bootandy: Not a problem! we all do it.

PR #401, opened 2024-05-27
Fix -x option behavior

rindeal:
This PR addresses an issue where `target_dirs` were being optimized out before being passed to the `get_filesystem_devices()` function. The changes in this PR ensure that `target_dirs` are passed directly to `get_filesystem_devices()`, and only then are they simplified.

Example of a command that currently doesn't work correctly:
```sh
dust -x $(findmnt -S tmpfs -o target -n)
```

It should show the usage of all tmpfs mounts, but it currently shows just the the root mountpoints like `/run`, since all the mountpoints nested under it are optimized out.

### DIFF
--- a/src/dir_walker.rs
+++ b/src/dir_walker.rs
@@ -111,7 +111,7 @@ fn ignore_file(entry: &DirEntry, walk_data: &WalkData) -> bool {
     let is_dot_file = entry.file_name().to_str().unwrap_or("").starts_with('.');
     let is_ignored_path = walk_data.ignore_directories.contains(&entry.path());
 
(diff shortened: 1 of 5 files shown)


bootandy:
Nice find, good edge case.

bootandy:
Fix the lint and I'll accept - thanks

bootandy:
Thanks, 

Agree with the streamlining. 

Those tests, were new tests and I refactored them here: https://github.com/bootandy/dust/pull/404 - so we no longer need that part of your PR. - which now causes a conflict :-(. 

(Although I do like your use of `$(mktemp -d)` so feel free to keep that if you'd like).

PR #405, opened 2024-06-19
feat: Add flag to read from file/stdin

Its-Just-Nans:
The `--files0-from=-` option of `du`

image

image


Its-Just-Nans:
Yes, you're rigth !

image

image



Note that this still hase an issue:

if someone do `dust --files0-from=-` without stdin, dust will block on the stdin

image

And you can leave with `CTRL+D` or kill the program

Its-Just-Nans:
Okay I made a fix using an `Arc<AtomicBool>` to check if we need to abort

src/main.rs
```diff
@@ -119,17 +122,49 @@ fn main() {
     let error_listen_for_ctrlc = Arc::new(Mutex::new(errors));
     let errors_for_rayon = error_listen_for_ctrlc.clone();
     let errors_final = error_listen_for_ctrlc.clone();
+    let is_in_listing = Arc::new(AtomicBool::new(false));
+    let cloned_is_in_listing = Arc::clone(&is_in_listing);
 
     ctrlc::set_handler(move || {
         error_listen_for_ctrlc.lock().unwrap().abort = true;
         println!("\nAborting");
+        if cloned_is_in_listing.load(Ordering::Relaxed) {
```
bootandy, line 131: I'm not sure we even need this boolean. If we detect a ctrl-c we should just abort irrespective of what `cloned_is_in_listin` is. 

How about we remove the 'if' but keep the process::exit(1) ?

bootandy: Do you think  'Aborting' should be eprintln instead of println? What do you think ?

Its-Just-Nans: I added the boolean check because if we just use process::exit(1), I'm not sure what would happen when we are in the Rayon threads. Nevertheless, I can remove it if needed.

For me, I think it's nice to have "Aborting" in the console to show that dust has correctly intercepted the signal.

bootandy:
ok, if you think people will use this, i'll include it.

PR #410, opened 2024-07-03
deps: update winapi-util to 0.1.8

ognevny:
starting with 0.1.7, winapi-util uses windows-rs, which is maintained official bindings from MS. this reduces dependence on winapi, which is abandoned for more than 3 years


bootandy:
thanks

Issue #421, opened 2024-07-26
du / dust disagreement


ammaraziz:
Hi Bootandy,

Thanks for creating and maintaining `dust` it is excellent piece of software.

I am finding that `du` disagrees with `dust` by several TB. Would you have any idea what could be causing the difference?

The commands I am running on the same directory:

```
sudo du -x -h --max-depth=1 
sudo dust -x -d 1
```

`du` agrees with `df` in regards to the total space used. The difference is something like 4TB of data. I have noticed that `dust` and `du` round numbers differently.

wugeer:
> Hi Bootandy,
> 
> Thanks for creating and maintaining `dust` it is excellent piece of software.
> 
> I am finding that `du` disagrees with `dust` by several TB. Would you have any idea what could be causing the difference?
> 
> The commands I am running on the same directory:
> 
> ```
> sudo du -x -h --max-depth=1 
> sudo dust -x -d 1
> ```
> 
> `du` agrees with `df` in regards to the total space used. The difference is something like 4TB of data. I have noticed that `dust` and `du` round numbers differently.

You can try adding the -s parameter.
![image](https://github.com/user-attachments/assets/f6bf9d20-578e-4949-9d8d-1260d5a05d9f)

wugeer:
> > Hi Bootandy,
> > Thanks for creating and maintaining `dust` it is excellent piece of software.
> > I am finding that `du` disagrees with `dust` by several TB. Would you have any idea what could be causing the difference?
> > The commands I am running on the same directory:
> > ```
> > sudo du -x -h --max-depth=1 
> > sudo dust -x -d 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `du` agrees with `df` in regards to the total space used. The difference is something like 4TB of data. I have noticed that `dust` and `du` round numbers differently.
> 
[...]

PR #422, opened 2024-07-28
fix: retrieve metadata for symbolic links without following them

wugeer:
# Background

Previously, the function get_metadata in platform.rs used `fs::metadata` which follows symbolic links and returns metadata for the target file. This caused issues https://github.com/bootandy/dust/issues/421: **du / dust disagreement**  when trying to determine properties of the symbolic link itself   

# Full Changelogs

- fix: retrieve metadata for symbolic links without following them

# Issue Reference

du / dust disagreement： https://github.com/bootandy/dust/issues/421

# Test Result

cargo test command output

image

du -s output vs cargo run -- -ok /home/shirley/code/python/official/superset/superset-frontend

image

du -s -L output vs cargo run -- -ok -L /home/shirley/code/python/official/superset/superset-frontend

image

while use dust v1.0.0 command,
image

image

The output shows that dust v1.0.0 produces the same results regardless of whether the -L parameter is used. That disagrees with du command.

# performance test results

image
image

image
image

### DIFF
--- a/src/dir_walker.rs
+++ b/src/dir_walker.rs
@@ -115,9 +115,11 @@ fn sort_by_inode(a: &Node, b: &Node) -> std::cmp::Ordering {
 fn ignore_file(entry: &DirEntry, walk_data: &WalkData) -> bool {
     let is_dot_file = entry.file_name().to_str().unwrap_or("").starts_with('.');
     let is_ignored_path = walk_data.ignore_directories.contains(&entry.path());
+    let follow_links =
+        walk_data.follow_links && entry.file_type().map_or(false, |ft| ft.is_symlink());
 
     if !walk_data.allowed_filesystems.is_empty() {
-        let size_inode_device = get_metadata(entry.path(), false);
+        let size_inode_device = get_metadata(entry.path(), false, follow_links);
         if let Some((_size, Some((_id, dev)), _gunk)) = size_inode_device {
             if !walk_data.allowed_filesystems.contains(&dev) {
                 return true;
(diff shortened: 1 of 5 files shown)


bootandy:
This doesn't seem to make a difference on my file system, but I'm happy to accept this.

wugeer:
> This doesn't seem to make a difference on my file system, but I'm happy to accept this.

On my two laptops, as long as there is a soft link file in a certain directory, and the target file of the soft link is much larger than the soft link itself, you can clearly see this difference. :)

bootandy:
I think it depends which filesystem you use. 

```
 df -T ~/
Filesystem     Type 1K-blocks      Used Available Use% Mounted on
/dev/nvme0n1p5 ext4 799971252 110782740 648520372  15% /

```

I'm using `ext4`. What are you using ?

wugeer:
> I think it depends which filesystem you use.
> 
> ```
>  df -T ~/
> Filesystem     Type 1K-blocks      Used Available Use% Mounted on
> /dev/nvme0n1p5 ext4 799971252 110782740 648520372  15% /
> ```
> 
> I'm using `ext4`. What are you using ?

Maybe you are right! it depends which filesystem you use. But i will take some time to set up an environment using the ext4 file system to see if I can reproduce the issue. :)
I'm using `btrfs`.
```
df -T ~/
Filesystem            Type 1K-blocks      Used       Available        Use%          Mounted on
/dev/nvme0n1p2 btrfs 999895364 716041748 273075852   73%         /home
```

wugeer:
Hi @bootandy 
I can confirm that this difference is related to the file system. The same file on my local laptop, which uses the btrfs file system, shows a block count of 8 when using the stat command, while on the comparison environment with the ext4 file system, the block count is 0. That's why this bug doesn't reproduce on the ext4 file system.
image
image

bootandy:
Nice. Good investigation. Well we have merged your PR now anyway.

PR #425, opened 2024-08-06
fix: 64-bit atomics for platforms with no 64-bit atomics

NoisyCoil:
`AtomicU64` cannot be used on platforms without 64-bit atomics, such as armel. This PR adds a dependency on the `portable-atomic` crate for those platforms, using the type of the same name defined in it. It introduces no changes for all other platforms (in particular, those on which dust currently builds are unchanged). I tested that this patch cross-builds to armel, and that all tests pass in an emulated armel environment.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,9 @@ sysinfo = "0.27"
 ctrlc = "3.4"
 chrono = "0.4"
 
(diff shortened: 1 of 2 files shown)


bootandy:
Looks harmless, 

Run `cargo fmt` and  `cargo clippy` and we should be good.

NoisyCoil:
Done, thanks!

Issue #430, opened 2024-08-15
Set up multiple invert-filter regex in config.toml


svm-zhang:
Hello,

Thank you for the awesome tool.

I have a question about setting up multiple invert filter REGEX in the config.toml file. I have checked the config example in the repo, as well as other tickets of similar matter. Could you please kindly show me an example?

I have also tried to add `reverse=true` and `iso=true` in the config.toml file, as shown in the config example. It seems that `reverse=true` gets cancelled by adding `iso=true`. Again, I likely did something wrong. Would be great if you could help me. Thanks again.

Simo

wugeer:
Hello, I will try to answer your question. 

Regarding your first question, currently, only the following parameter is supported in the `config.toml` file.
![image](https://github.com/user-attachments/assets/5aa1b781-5e98-4251-ba64-9a5806dd6ce9)

Regarding your second question, since now `iso` is not a valid configurable parameter, an error occurred when parsing the `config.toml` file, which caused it to fall back to using the default empty configuration. In this case, the actual configuration is determined by the runtime parameters. 

Now, this parameter has been replaced by `output-format="si"`.

So you can refer to this for your config.toml content.
![image](https://github.com/user-attachments/assets/0f5baeb0-5ee1-46aa-b821-3f39f99e64e4)

I will submit a PR to update this file.
![image](https://github.com/user-attachments/assets/2fe93b64-70cf-4334-a79d-b27f77507c1b)

svm-zhang:
@wugeer, much appreciate your help. It makes sense. Hopefully, we could specify multiple REGEXs in an array in the TOML file in the future. I will keep using alias for now.

bootandy:
Thank you @wugeer

svm-zhang:
@bootandy, could I expect this be implemented as a feature in the future? Thanks

bootandy:
I never thought people would want to put a regex filter in config, let alone multiple, but yes I think we can add that.

svm-zhang:
Thank you. I am learning Rust and if I feel comfortable with it in the near future, I will try to implement it and have a PR, given that this is likely a rare use case.

PR #431, opened 2024-08-16
Docs: update sample config.toml

wugeer:
Update `iso=true` in the config.toml to `output-format="si"` to fix issue: https://github.com/bootandy/dust/issues/430


bootandy:
Thank you. Yes.

Issue #444, opened 2024-10-14
Handle deleted files on Linux


tatref:
Hi,

On Linux, if a file is deleted while a process still has a handle on it, the disk space is still used, but not visible on the FS (ls, find.. will not find it). This is a common sysadmin issue, so I think it would be great to add an option to search for deleted files

The only way to know that a file is still taking up space, is by walking the file descriptors under `/proc/$pid/fd/`, and checking if the files still exist.
We can use `lsof` to show the deleted files:

```
[root@enterprise ~]# lsof -n | grep deleted
httpd    2357 apache   29u   REG 253,17 3926560     0  1499 /tmp/.NSPR-AFM-3457-9820130.0 (deleted)
mysqld   2588  mysql    4u   REG 253,17      52     0  1495 /tmp/ibY0cXCd (deleted)
mysqld   2588  mysql    5u   REG 253,17    1048     0  1496 /tmp/ibOrELhG (deleted)
```

Do you think it would be a worthy feature to add to dust?

bootandy:
That is an interesting edge case.

I can imagine this being a common problem for sysadmins.

dust works by walking thru the filesystem, if `ls` won't find it I don't think dust will. 


I'm currently not keen on adding this feature. Like you said, it would require walking the file descriptors, checking if that path matched the path dust was run for, then working out if it had already been included. Then if I were to show it I'd need a way of marking it as 'different' because it wouldn't be removed if you 'rm' the file. 

I think we are probably better served with `lsof -n | grep deleted`

tatref:
Yes you are correct, dust or ls can't find the file with a syscall in the filesystem, the only way is through /proc/

The workflow you describe is what I imagined. Yes `rm` can't delete the file, but the space is still used on the FS. We could maybe add a flag `--list-deleted` or something, then display the file deleted files as others. Or don't add a flag, and list the file with a different color/pattern

The thing is, using `lsof` can be complicated: the size shown is not the used space on the FS (it does not take into account sparse files), same file can be listed multiple time... Also there is no easy way of grouping the files by dir, as with dust.

I can help to make a demo implementation if you want

bootandy:
If you were to have this  `--list-deleted` flag. Do you think you would want it to be merged in with the regular files. Because I'm wondering if it should ONLY show the deleted files. 

If you are hunting down lost disk space your procedure for actual files in the filesystem is going to be different than for processes that are holding on to deleted files. 

So I'm proposing this:

```
dust --show-deleted 
[...]

PR #445, opened 2024-10-20
POC: deleted files

tatref:
Hi,

This is a quick and dirty POC for https://github.com/bootandy/dust/issues/444

Here are the changes

* I used [procfs](https://github.com/eminence/procfs) to walk the processes under `/proc/`, but we can do this manually if you don't want to add a new dependency. The function to list files is `get_deleted_files`
* I added a function to modify the node tree: `insert_deleted_file_in_node`, maybe it could be merged with `build_node`
* The handling of the deleted files is done in `walk_it`
* I also made `InodeAndDevice` `pub(crate)` and used it to replace a few `(u64, u64)`

The drawbacks:

* There is an edge case with multiple deleted hardlinks with different paths: in this case, only the 1st file will be listed
* I only handle the file size for now (no filecount, filetime...), also `WalkData` is not used (no regex, ignore_dir...)
* I hardcoded the flag to handle the deleted files

On the bright side, there is no change (performance or inner working) if we set `handle_deleted_files` to false

Let me know what you think about it

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ directories = "4"
 sysinfo = "0.27"
 ctrlc = "3.4"
 chrono = "0.4"
+procfs = { version = "0.17.0", default-features = false }
 
 [target.'cfg(not(target_has_atomic = "64"))'.dependencies]
 portable-atomic = "1.4"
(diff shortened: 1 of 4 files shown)


bootandy:
thanks, that's an interesting demonstration.

Issue #448, opened 2024-11-06
Directories greater than 1000TB do not pretty format nicely


janbridley:
Directories (or subdirectories) that contain greater than one petabyte of data do not pretty format as they could. I have a simple solution (and test) for this, but do not have access to push a PR - should I fork and merge from there?

Before:

```
╰─➤  dust -n 1
4729T   ┌── workspace│██████████████████████████████████████████████████ │  99%
4753T ┌─┴ .          │██████████████████████████████████████████████████ │ 100%
```

After fix:

```
╰─➤  dust -n 1
4.7P   ┌── workspace│██████████████████████████████████████████████████ │  99%
4.7P ┌─┴ .          │██████████████████████████████████████████████████ │ 100%
```

PR #449, opened 2024-11-07
Enable pretty format for petabyte data

janbridley:
As in #448, volumes greater than 1000TB do not pretty format. This has been addressed, and additional tests have been added for large folders.


bootandy:
looks good, lets format it and merge

PR #459, opened 2025-01-08
Unify long option names

pkulev:
Some of the long option names were separated with underscores.


bootandy:
You are right, and I will accept this. But I'm going to wait until a major release because this is a functional change.

bootandy:
https://github.com/bootandy/dust/pull/484
replaced with ^

Issue #452, opened 2024-12-05
Option (CLI and Config) to "collapse" output about children of certain directories


edbrannin:
Sometimes there's a directory whose size I want to see, but whose children I don't care about.

Examples:
- `.git`
    - (I almost never care about what's inside here, but knowing "This project's `.git` repo is 80% of the whole directory" would be helpful)
- `node_modules`
    - (Maybe I want to `rm -rf node_modules` in a stale project, but I usually don't care which dependencies are big)
- `target` in Maven projects
    - (Maybe I want to run `mvn clean` in some projects I haven't touched in months)
- `.next/cache` as seen below

It would be heopful if I could set a list of directories to keep collapsed in the config file, rather than building up an alias[1] in my `.zshrc`

[1] Something like: `alias dust=dust --collapse=.git --collapse=node_nodules --collapse=target --collapse=.next/cache` 

Example output where this would be helpful:

```bash
bruno % dust
 63M           ┌── lib                             │███░░░░░░░░░░░░░░░░░░░░░                                          │   3%
 64M         ┌─┴ typescript                        │███░░░░░░░░░░░░░░░░░░░░░                                          │   3%
 64M       ┌─┴ node_modules                        │███░░░░░░░░░░░░░░░░░░░░░                                          │   3%
 64M     ┌─┴ bruno-query                           │███░░░░░░░░░░░░░░░░░░░░░                                          │   3%
 63M     │     ┌── lib                             │███░░░░░░░░░░░░░░░░░░░░░                                          │   3%
 64M     │   ┌─┴ typescript                        │███░░░░░░░░░░░░░░░░░░░░░                                          │   3%
 64M     │ ┌─┴ node_modules                        │███░░░░░░░░░░░░░░░░░░░░░                                          │   3%
 64M     ├─┴ bruno-common                          │███░░░░░░░░░░░░░░░░░░░░░                                          │   3%
 63M     │     ┌── lib                             │███░░░░░░░░░░░░░░░░░░░░░                                          │   3%
 64M     │   ┌─┴ typescript                        │███░░░░░░░░░░░░░░░░░░░░░                                          │   3%
 64M     │ ┌─┴ node_modules                        │███░░░░░░░░░░░░░░░░░░░░░                                          │   3%
 64M     ├─┴ bruno-graphql-docs                    │███░░░░░░░░░░░░░░░░░░░░░                                          │   3%
[...]

PR #466, opened 2025-01-26
feat: New --collapse flag

bootandy:
--collapse will keep that directory collapsed and will not expand it.


bootandy:
For: https://github.com/bootandy/dust/issues/452

Issue #441, opened 2024-10-01
Improve output when used with a wildcard expression


Debilski:
When doing something like `dust /home/*/.local/share/Trash` the output looks like this, which is somewhat suboptimal:

```
 11G   ┌── Trash     │███                                                                        │   2%                                                                                           
 12G   │ ┌── files   │███                                                                        │   2%                                                                                           
 12G   ├─┴ Trash     │███                                                                        │   2%                                                                                           
 17G   │ ┌── files   │█████                                                                      │   3%                                                                                           
 17G   ├─┴ Trash     │█████                                                                      │   3%                                                                                           
...
300G ┌─┴ (total)     │██████████████████████████████████████████████████████████████████████████ │ 100%

```

 I think it would be nice to have an option that included the ‘minimal path difference’ in the output, ie. in this case the name of the home folder.

I have seen the `-p`/`--full-paths` option but I would only need to have it in the first level. Even better, if it was just shown as `Trash (user123)` in the output once.

bootandy:
Oh I see, you are looking at different user's Trash directories. Hmm, that's a tricky one, I'll think about this.

bootandy:
` Trash (user123)`
I suppose it would be possible to generate the output if we did a full diff on the path of the duplicate folders. 

Interesting.

PR #470, opened 2025-02-05
feat: Handle duplicate dir names better

bootandy:
If we run `dust /usr/*/Trash`
We see several 'Trash' directories in the output but do not know which user they belong to.

This fix means if we see duplicate names in a directory we will display the parent directory name as well

for: https://github.com/bootandy/dust/issues/441

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -1,3 +1,6 @@
+use stfu8::encode_u8;
+
+use crate::display::get_printable_name;
 use crate::display_node::DisplayNode;
 use crate::node::FileTime;
 use crate::node::Node;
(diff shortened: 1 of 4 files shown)


src/filter.rs
```diff
@@ -161,3 +169,56 @@ fn build_display_node(mut new_children: Vec<DisplayNode>, current: &Node) -> Dis
...
+                    }
+                }
+            }
+            new_top_nodes = newer;
+        }
+        new_top_nodes
+    } else {
+        top_level_nodes
+    }
+}
```
bootandy, line 224: add a test for this using below dup_name folders

Issue #471, opened 2025-02-12
-X option does not work for absolute paths


sataset:
# Info
- OS: nixos
- Dust version: 1.1.1

# Command 
```sh
cd /
sudo dust -X /mnt 
```

# Expected 
Ignore `/mnt` during calculation.

# Actual
`/mnt` not ignored.

<details><summary>Screenshots</summary>

![Image](https://github.com/user-attachments/assets/89c0fec6-b75c-48fa-b3dd-c393f89b286a)
![Image](https://github.com/user-attachments/assets/724f51c0-203b-4e28-a80d-0d481fc62f8e)
</details> 

# Workaround
Use relative path.

<details><summary>Screenshots</summary>

![Image](https://github.com/user-attachments/assets/96de5744-c957-4d04-9aee-2cd4eec79020)
</details>

bootandy:
True, -X doesn't work on absolute paths. 

$ /home/andy/
dust -X /home/andy/trash

Will still show the trash directory. 


I'll look if I can fix this. If not I'll update the docs.

PR #475, opened 2025-03-06
fix: Support absolute path with `-X` option

frendsick:
Allow excluding any directory or file by absolute path using the `-X` option.

The proposed change for the `ignore_file` function checks whether the file's absolute path representation starts with any of the absolute path representations of ignored directories. Any file inside an ignored folder should have a common path with the ignored directory.

Fixes: #471

### DIFF
--- a/src/dir_walker.rs
+++ b/src/dir_walker.rs
@@ -132,19 +132,14 @@ fn is_ignored_path(path: &Path, walk_data: &WalkData) -> bool {
     }
 
     // Entry is inside an ignored absolute path
+    // Absolute paths should be canonicalized before being added to `WalkData.ignore_directories`
     for ignored_path in walk_data.ignore_directories.iter() {
         if !ignored_path.is_absolute() {
             continue;
         }
-        match std::fs::canonicalize(ignored_path) {
-            Ok(absolute_ignored_path) => {
-                let absolute_entry_path =
-                    std::fs::canonicalize(path).unwrap_or_default();
-                if absolute_entry_path.starts_with(absolute_ignored_path) {
-                    return true;
-                }
-            }
-            Err(_) => continue,
+        let absolute_entry_path = std::fs::canonicalize(path).unwrap_or_default();
+        if absolute_entry_path.starts_with(ignored_path) {
+            return true;
         }
     }
 
--- a/src/dir_walker.rs
+++ b/src/dir_walker.rs
@@ -177,6 +177,9 @@ fn ignore_file(entry: &DirEntry, walk_data: &WalkData) -> bool {
 
     // Entry is inside an ignored absolute path
     for ignored_path in walk_data.ignore_directories.iter() {
+        if !ignored_path.is_absolute() {
+            continue;
+        }
         match std::fs::canonicalize(ignored_path) {
             Ok(absolute_ignored_path) => {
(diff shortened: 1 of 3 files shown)


frendsick:
I tested the changes' performance with Ubuntu 22.04 running on WSL. The performance hit is negligible when `dust` is run without the `-X` flag or when the path in the `-X` flag is relative. However, using absolute paths with `-X` flag is slower as there is some overhead from [std::fs::canonicalize](https://doc.rust-lang.org/std/fs/fn.canonicalize.html).

These are measurements for listing `/mnt/c` within my WSL using absolute and relative paths. The commands were run from the `/wsl/c` folder. According to these measurements, using an absolute path with the `-X` option is roughly 60% slower than using a relative path.

**Relative path (version 1.1.2)**

```
time dust . -X Users
...
real    1m27.669s
user    0m2.717s
sys     1m7.911s
```

**Relative path (new version)**

```
time ~/dust-fork/target/release/dust . -X Users
...
real    1m20.461s
user    0m2.375s
sys     0m59.442s
```

**Absolute path (new version)**

```
time ~/dust-fork/target/release/dust . -X /mnt/c/Users
...
real    2m11.157s
user    0m6.415s
sys     2m2.055s
```

frendsick:
If we would like to get rid of the overhead of calling [std::fs::canonicalize](https://doc.rust-lang.org/std/fs/fn.canonicalize.html) for the ignored path over and over, we could probably do the necessary `canonicalize` call for absolute ignored paths before creating [walk_data](https://github.com/bootandy/dust/blob/d876cc28a769ead9498497411a0feb1be6c02370/src/main.rs#L264). One spot could be when constructing [ignored_full_paths](https://github.com/bootandy/dust/blob/d876cc28a769ead9498497411a0feb1be6c02370/src/main.rs#L233). Then we would still need to canonicalize [&entry.path()](https://github.com/bootandy/dust/blob/a1638684b26b1be8f145be832b51b42a5aedb304/src/dir_walker.rs#L186). Still, there is no need to canonicalize the [ignored_path](https://github.com/bootandy/dust/blob/a1638684b26b1be8f145be832b51b42a5aedb304/src/dir_walker.rs#L183) everytime we want to check if an entry should be ignored.

If this were implemented, the overhead penalty for checking absolute paths should be cut in half.

bootandy:
thanks for taking a look at this - I've started to play with this idea on a local branch.

src/dir_walker.rs
```diff
@@ -175,6 +175,23 @@ fn ignore_file(entry: &DirEntry, walk_data: &WalkData) -> bool {
         return true;
     }
 
+    // Entry is inside an ignored absolute path
```
bootandy, line 178: Let's pull this function out to a separate method and combine it with the call on line 130:

`let is_ignored_path = walk_data.ignore_directories.contains(&entry.path());`

bootandy:
I like this - I think this will work- thanks

bootandy:
> `ignored_full_paths`

I've read this comment after my review - I think you should make those changes.

frendsick:
Thank you for the review. I will make the requested changes and fix the clippy warnings, possibly later today.

bootandy:
looks good! thanks

PR #478, opened 2025-03-16
feat: Abort immedietly when ^C is received.

### LINKED ISSUE #477: ^C fails to immediately abort indexing
This is an issue very similar to https://github.com/bootandy/dust/issues/367, though I'm opening a separate issue because the issue manifests in a slightly different way.

### Expected Behavior
When I send a SIGINT signal to a running `dust` process (via `kill` or `C-c`), I expect the process to quit immedietly.

### Actual Behavior
When indexing a large directory, or indexing on a slow filesystem, attempting to interrupt the process causes `dust` to continue processing, sometimes for a long period of time. I am able to kill it via SIGKILL, though intuitively, I might prefer it to handle a `C-c` in a way that immediately kills the process.

#### Example Reproduction
```
Indexing: ../my_directory 1596252 files, 6.6T ... \^C
Aborting
Indexing: ../my_directory 1604186 files, 6.6T ... /^C
Aborting
Indexing: ../my_directory 1672889 files, 6.6T ... -^C
Aborting
Indexing: ../my_directory 1947295 files, 6.7T ... |^C
Aborting
Indexing: ../my_directory 2415938 files, 6.9T ... \Killed
<process exits>
```

In this reproduction, I attempt to exit the running process with `C-c` several times, which lets me know that the process is "Aborting", but continues running. Finally, I kill the process with `kill -9 <PID>`, at which point the process exits.

### Investigation
I am able to consistently reproduce this issue on the most recent commit as of the posting of this issue: `dd799706fbd062b468d5e8e219c0ef332344588b`.

It seems that the [SIGINT](https://github.com/bootandy/dust/blob/dd799706fbd062b468d5e8e219c0ef332344588b/src/main.rs#L129) handler configured attempts to perform a graceful shutdown when indexing is in process, choosing to set `error_listen_for_ctrlc.abort` rather than immediately exiting.

This flag is checked during [directory walking](https://github.com/bootandy/dust/blob/dd799706fbd062b468d5e8e219c0ef332344588b/src/dir_walker.rs#L209), but only at the very beginning of a walk for a directory -- if a walk for a large directory (or a directory on a very slow filesystem) is already in progress, it will wait until the walk completes, which may take a significant period of time.

### Suggestions
I'm not entirely certain why we set `error_listen_for_ctrlc.abort` when we receive SIGINT, rather than simply exiting the process immediately. Unless there's some behavior I'm not aware of we're attempting to handle, it seems to me that we could rewrite this function as follows:

```
diff --git a/src/main.rs b/src/main.rs
index 0a976eb..79def97 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -126,9 +126,7 @@ fn main() {
     ctrlc::set_handler(move || {
         error_listen_for_ctrlc.lock().unwrap().abort = true;
         println!("\nAborting");
-        if cloned_is_in_listing.load(Ordering::Relaxed) {
-            process::exit(1);
-        }
+        process::exit(1);
     })
     .expect("Error setting Ctrl-C handler");
```

When running with this patch applied, I get the behavior I expect -- namely, an exit as soon as SIGINT is received:

```
Indexing: ../my_directory 1140107 files, 6.4T ... -^C
Aborting
<process exits>
```

If we _do_ want to handle SIGINT gracefully for some reason I'm not aware of, I might suggest some sort of flag that forcefully exits the process on the second SIGINT, while displaying a message during the first that we are attempting to shut down gracefully and to press `C-c` for immediate abort (like the following, approximately):

```
Indexing: ../my_directory 1596252 files, 6.6T ... \^C
Aborting indexing, press C-c again to force immediate abort
Indexing: ../my_directory 1740107 files, 7.4T ... -^C
Forcefully aborting now
<process exits>
```

Alternatively, some sort of command-line option that immediately aborts rather than attempting to finalize indexing may be a viable option.

I'm happy to put out a PR for any of these changes (though in my personal opinion, the first approach -- immediate exit -- meshes best with how I would naively expect the program to behave).

Thanks for the great tool! I've gotten a lot of great usage out of it 😄

quantatic:
Previously, we attempted to perform a clean shutdown, which could take a significant period of time on slow filesystems. This commit changes the shutdown logic to unconditionally abort immediately when ^C is received by the program.

This PR resolves #477.

### DIFF
--- a/src/dir_walker.rs
+++ b/src/dir_walker.rs
@@ -206,10 +206,6 @@ fn walk(dir: PathBuf, walk_data: &WalkData, depth: usize) -> Option<Node> {
     let prog_data = &walk_data.progress_data;
     let errors = &walk_data.errors;
 
-    if errors.lock().unwrap().abort {
-        return None;
-    }
-
     let children = if dir.is_dir() {
         let read_dir = fs::read_dir(&dir);
         match read_dir {
--- a/src/main.rs
+++ b/src/main.rs
@@ -26,8 +26,6 @@ use std::panic;
 use std::process;
 use std::sync::Arc;
 use std::sync::Mutex;
-use std::sync::atomic::AtomicBool;
-use std::sync::atomic::Ordering;
 use sysinfo::{System, SystemExt};
 use utils::canonicalize_absolute_path;
 
@@ -119,20 +117,13 @@ fn main() {
     let errors = RuntimeErrors::default();
     let error_listen_for_ctrlc = Arc::new(Mutex::new(errors));
     let errors_for_rayon = error_listen_for_ctrlc.clone();
-    let errors_final = error_listen_for_ctrlc.clone();
-    let is_in_listing = Arc::new(AtomicBool::new(false));
-    let cloned_is_in_listing = Arc::clone(&is_in_listing);
 
     ctrlc::set_handler(move || {
-        error_listen_for_ctrlc.lock().unwrap().abort = true;
         println!("\nAborting");
-        if cloned_is_in_listing.load(Ordering::Relaxed) {
-            process::exit(1);
-        }
+        process::exit(1);
     })
     .expect("Error setting Ctrl-C handler");
 
-    is_in_listing.store(true, Ordering::Relaxed);
     let target_dirs = match config.get_files_from(&options) {
         Some(path) => {
             if path == "-" {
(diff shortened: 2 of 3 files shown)


quantatic:
This PR _may_ resolve #478, though we'll probably need confirmation from the original reporter of that issue.

PR #481, opened 2025-03-30
Less `fs::metadata` call in `fn walk(`

ldm0:
`PathBuf::is_dir`/`PathBuf::is_file` calls `fs::metadata` individually and can be squashed.


bootandy:
I can see you've removed a call to `fs::metadata` here which is good. 

But the extra code makes it slightly harder to read as well. 

Testing with hyperfine, I can't see any performance improvement with this version so I suspect the call to `fs::metadata` is cached. 

I'm not convinced this is worth adding as it doesn't seem to have a performance improvement. 

Good idea though.

PR #493, opened 2025-05-15
chore(cli): Migrate the definition of CLI to the Derive API

sorairolake:
Change the definition of CLI from the [Builder API](https://docs.rs/clap/4.5.38/clap/_tutorial/index.html) to the [Derive API](https://docs.rs/clap/4.5.38/clap/_derive/_tutorial/index.html).

Also add the help message for the positional argument and the possible values.


bootandy:
this looks good, will look at closely soon

Issue #495, opened 2025-05-23
Dust never finishes


sasq64:
Did `cargo install` and ran it in my home directory.
Quickly counted up to

`Indexing: . 2835562 files, 299G ... /`

but then doesn't seem to get any further. Interrupted it after 5 minutes.

Amr-Shams:
could u use `dust --print-errors`

sankalp-khare:
Same issue on `v1.2.0` (installed using homebrew).
Tried with `dust --print-errors` but it kept spinning and no error details got printed.

I built `v1.1.1` locally and ran it on the same filesystem, and the issue did not happen.

Looks like a regression introduced in `v1.2.0`.

Is there a way I could get dust to print what it is doing? I don't see a `--verbose` equivalent in the flags.

sasq64:
My suspicion was that it tried to access a folder that it didn't have permissions for, because I was using a new terminal emulator on macOS when I got the error...

sankalp-khare:
@Amr-Shams thanks for your commit adding `-V` option!
Using the release executable built from https://github.com/Amr-Shams/dust/commit/419e6699785c0b2b71434946988a4cd64d436eec I can see the issue:

```
Scanning: ./Library/Containers/com.readitlater.PocketMac/Data/Library/Application Support/Pocket/offline/cache0/RIL_assets/1.bp.blogspot.com/-McqWCErHM6s/XADb1-1fVqI/AAAAAAAAMsY/9u0U361v_MMHaVVPZfwBhm3jEFajYQPBACK4BGAYYCw/s1600
    File: ./Library/Containers/com.readitlater.PocketMac/Data/Library/Application Support/Pocket/offline/cache0/RIL_assets/1.bp.blogspot.com/-McqWCErHM6s/XADb1-1fVqI/AAAAAAAAMsY/9u0U361v_MMHaVVPZfwBhm3jEFajYQPBACK4BGAYYCw/s1600/Slide003.png (size: 176128)
Indexing: . 4586481 files, 287G ... |Error reading directory ./Library/Containers/com.readitlater.PocketMac/Data/Library/Application Support/Pocket/offline/cache0/RIL_assets/th-thumbnailer.cdn-si-edu.com/ztcTnL0_Q_OJVCbNLuuU9Lc_r3A: Interrupted system call (os error 4)
Indexing: . 4586481 files, 287G ... -
```

is where it's getting stuck. I am, however, able to traverse the directory using `find` without seeing any error:
```
[...]

PR #497, opened 2025-06-02
fix: bug: Remove bad error handling

bootandy:
This may be causing dust to lock up as we were accidentally creating a second editable_error causing threads to lock.

https://github.com/bootandy/dust/issues/495


sankalp-khare:
@bootandy I see there's a bunch of checks run against Pull Request branches. Wonder if we could setup a test case or something which would detect this kind of issue.

Although even for the dir tree in my filesystem where the locking occurred, it happened non-deterministically... Sometimes a `dust` run for the same dir tree would complete, other times it'd encounter this. So I'm not very sure what the test case / job would look like.

PR #533, opened 2025-11-07
Update directories from 4 to 6

musicinmybrain:
Update the `directories` dependency to the latest and final version, 6.0.0.

The SemVer-breaking change in `directories` 6 is just that `dirs-sys` is updated to 0.5, which, in turn, is just due to [updating some platform-specific dependencies](https://github.com/dirs-dev/dirs-sys-rs/compare/f369f0904bec9833572f24c988c7e48454173983...8bcd4aa2c35990d57a2cff2953793525fc42709c). There are therefore no API changes we need to worry about here. The update from `directories` 4 to 5 was similar: https://github.com/dirs-dev/directories-rs/tree/4d76f1a5a0a979de3cac4c1ffb53c8b5a6f481a7?tab=readme-ov-file#6

Note that the [repository for `directories` is archived](https://github.com/dirs-dev/directories-rs), so it might be time to start thinking about a maintained alternative.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -237,7 +237,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "df51e72c150781d2c7d4cbcb0b803277caaa80476786994a62961a8f1010dafb"
 dependencies = [
  "serde",
- "thiserror",
+ "thiserror 1.0.69",
  "toml",
 ]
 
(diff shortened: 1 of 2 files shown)


bootandy:
directories is used in 1 place! I'll look at removing it.

bootandy:
https://github.com/bootandy/dust/pull/538

PR #568, opened 2026-04-02
fix: prevent panic on repeated EINTR errors during filesystem scanning

hieuit095:
When dust scans filesystems that return many EINTR errors (e.g., network filesystems like NFS or FUSE-based mounts), the previous code called panic after 999 consecutive interrupts, crashing the program. Replaced the panic with a warning message on stderr and a graceful skip of the problematic directory, allowing the scan to continue with remaining accessible paths.

Changes: src/dir_walker.rs - in handle_error_and_retry(), panic replaced with eprintln warning and return false when interrupted_error exceeds 999.

All 29 unit tests and 28 integration tests pass.


bootandy:
Oh wow that actually happened? I put 999 as an abstract 'it cant possible get here number'.

bootandy:
I'm happy to include this if it helps.

PR #571, opened 2026-04-17
fix: use IEC prefixes (Ki, Mi, Gi) for base-1024 size display

### LINKED ISSUE #570: dust shows IEC values with SI prefixes
The manual shows these options for `--output-format`:
- `si`
- `b`
- `k` - `t` for `KiB` - `TiB` (IEC)
- `kb` - `tb` for `KB` - `TB` (SI)

To me, this implies that it defaults to SI prefixes, and the output using `M` or `G` suggests the same. However the reported sizes are actually using IEC prefixes and are "larger" when specifying `-o si`.

Testing with these files:
```
$ fallocate -l 1000000000 1GB.bin
$ fallocate -l 1G 1GiB.bin
$ fallocate -l 1000000 1MB.bin
$ fallocate -l 1M 1MiB.bin
```

`eza` shows the correct units:
```
$ eza -l --no-permissions --no-user --no-time (default is SI)
1.0G 1GB.bin
1.1G 1GiB.bin
1.0M 1MB.bin
1.0M 1MiB.bin

$ eza -l --no-permissions --no-user --no-time -b (for IEC prefixes)
954Mi 1GB.bin
1.0Gi 1GiB.bin
977Ki 1MB.bin
1.0Mi 1MiB.bin

$ eza -l --no-permissions --no-user --no-time -B (for plain byte count)
1,000,000,000 1GB.bin
1,073,741,824 1GiB.bin
    1,000,000 1MB.bin
    1,048,576 1MiB.bin
```
Whereas `dust` doesn't differentiate between SI and IEC:
```
$ dust -bo si
1.0M   ┌── 1MB.bin
1.0M   ├── 1MiB.bin
1.0G   ├── 1GB.bin
1.1G   ├── 1GiB.bin
2.1G ┌─┴ .

$ dust -b
980K   ┌── 1MB.bin
1.0M   ├── 1MiB.bin
953M   ├── 1GB.bin
1.0G   ├── 1GiB.bin
1.9G ┌─┴ .

$ dust -bo b
   1003520B   ┌── 1MB.bin
   1048576B   ├── 1MiB.bin
1000001536B   ├── 1GB.bin
1073741824B   ├── 1GiB.bin
2075795456B ┌─┴ .
```
(Idk why `dust` reads a wrong size for the SI files, `eza`, GNU `ls`, and GNU `du` don't do this)

When printing the sizes with IEC prefixes, it should be `Ki`/`Mi`/`Gi`/`Ti`. The man page should also mention that it defaults to IEC prefixes.

ChihebBENCHEIKH1:
Closes #570

dust calculates sizes using base-1024 (binary/IEC) by default but displays them with SI prefixes (K, M, G), which is misleading since SI prefixes imply powers of 1000.

**Before:** `1.0M` (ambiguous — is this 1,000,000 or 1,048,576 bytes?)
**After:** `1.0Mi` (clearly 1,048,576 bytes, base-1024)

SI mode (`-o si`) continues using K, M, G with base-1000 as expected.

**Changes:**
- Replaced single `UNITS` array with `SI_UNITS` and `IEC_UNITS`
- `human_readable_number()` and `get_number_format()` now select the correct suffix based on the output format
- Updated all unit and integration tests

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -16,7 +16,8 @@ use std::iter::repeat_n;
 use std::path::Path;
 use thousands::Separable;
 
-pub static UNITS: [char; 5] = ['P', 'T', 'G', 'M', 'K'];
+pub static SI_UNITS: [&str; 5] = ["P", "T", "G", "M", "K"];
+pub static IEC_UNITS: [&str; 5] = ["Pi", "Ti", "Gi", "Mi", "Ki"];
 static BLOCKS: [char; 5] = ['█', '▓', '▒', '░', ' '];
 const FILETIME_SHOW_LENGTH: usize = 19;
 
@@ -412,6 +413,14 @@ fn get_pretty_name(
     }
 }
 
+pub fn get_units(output_str: &str) -> &'static [&'static str; 5] {
+    if get_type_of_thousand(output_str) == 1024 {
+        &IEC_UNITS
+    } else {
+        &SI_UNITS
+    }
+}
+
 // If we are working with SI units or not
 pub fn get_type_of_thousand(output_str: &str) -> u64 {
     if output_str.is_empty() {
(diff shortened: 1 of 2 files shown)


bootandy:
it was originally designed this way to save screen real-estate. But this does make things clearer, so I'm accepting.

PR #572, opened 2026-04-22
feat: add --dim flag to render percent bars in dark grey

aleeraser:
This has been bugging me since quite some time.  
Bars default to terminal color; --dim applies DarkGray (ANSI 8) to reduce brightness on dark-background terminals.  
Also configurable via `dim = true` in config.toml.

Note: this is AI generated, but built and tested manually.

Before:

image

After:

image


bootandy:
I think this is a reasonable request. I'm now wondering if dim should be the default.

aleeraser:
Well, in that case the whole PR can become a one line change :)

bootandy:
yeah, I'm going to think about that, I'll either accept this or do a one line change.

bootandy:
https://github.com/bootandy/dust/pull/595
fixed the ci build here ^

PR #578, opened 2026-05-27
add --output-metrics format using prometheus-client

clonejo:
I wanted to see over time where my disk space was being lost to, so i now run this a loop. Example output:

```console
`--> cargo run -- --output-metrics src target -n 10
   Compiling du-dust v1.2.4 (/var/home/clonejo/code/dust)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.48s
     Running `target/debug/dust --output-metrics src target -n 10`
# HELP dust_file_size_bytes Total size of files in this folder / size of this file.
# TYPE dust_file_size_bytes gauge
# UNIT dust_file_size_bytes bytes
dust_file_size_bytes{paths="src target",path="(total)"} 1004326912
dust_file_size_bytes{paths="src target",path="target"} 1004191744
dust_file_size_bytes{paths="src target",path="target/debug"} 1003950080
dust_file_size_bytes{paths="src target",path="target/debug/deps"} 622583808
dust_file_size_bytes{paths="src target",path="target/debug/deps/dust-f39119bf77cf848e"} 48771072
dust_file_size_bytes{paths="src target",path="target/debug/deps/dust-1ba6c2ffc2c7adf6"} 48517120
dust_file_size_bytes{paths="src target",path="target/debug/deps/dust-513f5be0821b1cdc"} 47734784
dust_file_size_bytes{paths="src target",path="target/debug/incremental"} 277671936
dust_file_size_bytes{paths="src target",path="target/debug/incremental/dust-2ede6adlvlb1p"} 50765824
dust_file_size_bytes{paths="src target",path="target/debug/incremental/dust-117p7tduc8h0u"} 50200576
dust_file_size_bytes{paths="src target",path="target/debug/build"} 99995648
# EOF
```
This can then be consumed by [node-exporter](https://github.com/prometheus/node_exporter)'s `textfile` collector.

I had originally hand-rolled the prometheus text format, but by adding a dependency generating the format correctly is already solved.

What is your opinion on adding such a feature?

Obvs this still needs tests, and the added code in `display_node.rs` is probably big enough to move into a separate module.


bootandy:
I'm not keen on flags that are only used for specific products, and that's quite a lot of extra stuff in DisplayNode  so I'm going to decline.

thanks for your thoughts though.

PR #583, opened 2026-06-20
fix: don't request full system info on startup

DaniPopes:
`init_rayon` is calling `System::new_all` which, at least on Linux, parses every single entry in /proc, issuing multiple syscalls per entry. On my system this currently amounts to ~0.5s of overhead on any invocation of `dust`.

Use `System::new` since `init_rayon` only needs `refresh_memory`.


arcuru:
Just to cross-link, this fix is also in #574

PR #587, opened 2026-07-24
feat: add `limit-filesystem` config file option

stevenwcarter:
Most configuration options that could be specified on the command line existed in the config file, but `limit-filesystem` was absent. Mirrors the existing patterns to add this support, along with minimal documentation changes. Adds a helpful link for new users to the sample config file as well in the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,9 @@ Either: `~/.config/dust/config.toml` or `~/.dust.toml`
 ```
 $ cat ~/.config/dust/config.toml
 reverse=true
+limit-filesystem=true
 ```
+Keys use the long flag name in kebab-case. See [config/config.toml](config/config.toml) for a fuller sample.
 
 ## Alternatives
 
--- a/config/config.toml
+++ b/config/config.toml
@@ -24,6 +24,9 @@ skip-total=true
 # Do not display hidden files
 ignore-hidden=true
 
+# Only count files and directories on the same filesystem as the supplied directory
+limit-filesystem=true
+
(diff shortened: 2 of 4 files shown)


bootandy:
true there isn't a config option for that, but is there a situation when we would want that in config ?

stevenwcarter:
I almost never want to have du/dust report on the sizes of my NFS/CIFS mounted folders. I have terabytes worth of network storage mounted via autofs in my home directory so I was looking for an option to disable it by default. It was either this, or setting up an alias to add the command-line flag by default. I use these tools to figure out which directories are taking up my disk space, so I generally only want it to report on the current disk.

Everyone's usage varies, and it's pretty straightforward to enable it as an option. Given `dust` already supports a config file, and the option is already a command-line flag, I don't think it would be hurting future maintainability to include it for those who might want it as the default behavior like me.

Arguably it would be better if it could limit to local disk storage and make network storage an opt-in behavior, but I don't know if there are clean cross-platform ways to do that reliably.

stevenwcarter:
~Well after an embarrassingly small amount of research, it seems that https://lib.rs/crates/inspect_path would fit this use-case perfectly. If you'd prefer, I could add a separate configuration option to include/exclude remote directories and wire this up. It would mean an additional dependency, but would be a better overall option if you prefer.~

~My personal needs are met with just the change in this PR, but I don't mind taking the time to wire this other crate up if you'd prefer.~

That crate may not support mac.

bootandy:
Yeah ok I'll accept.

PR #588, opened 2026-07-25
feat: add dotted padding line between file/folder name and bar

stevenwcarter:
It could sometimes be hard to tell which of two lines was the one where a large jump would occur in their size on the bar chart. This was especially problematic when they were much shorter than other filenames, leading to a very large gap between the names and the bar chart.


stevenwcarter:
Screenshot of new appearance for reference:

image

bootandy:
I've wondered about this before.  - What about having dots at the end of the bar graph too? Or in the bar graph? 

I'm going to say no for now. - I'm trying to limit number of flags. 

But I might change the gui someday so that '.' replaces the space.

PR #599, opened 2026-08-14
fix: don't truncate JSON output to the terminal height

### LINKED ISSUE #597: Feature request: Default to unlimited nodes when outputting JSON (-j / --output-json) or add --no-limit flag
## Feature Request

When running `dust` with `-j` / `--output-json` (e.g., in CI/CD pipelines, background scripts, or automated tools), `dust` currently defaults `-n` (`--number-of-lines`) to `terminal_height` (~20–30 lines).

Because JSON output (`-j`) is intended for machine parsing rather than human terminal TUI rendering, truncating the JSON tree to ~20 lines by default leads to incomplete JSON representations unless users pass arbitrary large parameters like `-d 1000 -n 10000000`.

## Proposed Solutions

1. **Default JSON output (`-j`) to unlimited depth/nodes**:
   When `--output-json` (`-j`) is active, `dust` could automatically default `number_of_lines` and `depth` to unlimited so that the output contains the complete directory tree by default.
   
2. **Add a `--no-limit` / `--all` flag**:
   Provide an explicit CLI flag (e.g., `--no-limit`) to easily request an un-truncated full tree traversal without needing to pass large numbers like `-n 10000000`.

## Context & Use Case
In automated CI environments (such as GitHub Actions or server disk profiling workflows), users invoke `dust -j /` expecting a full machine-readable JSON report of the disk hierarchy. Having `-j` constrained by TUI `terminal_height` causes surprising truncation.

https://github.com/Malix-Labs/GitHub-Action_Runner-Fetch/ is an example
see https://github.com/Malix-Labs/GitHub-Action_Runner-Fetch/blob/7655ac20d0537c747e0a36e4f24d4edc91fd35ae/src/fetch.sh#L100

Thanks for maintaining `dust`!

VXNCXNX:
Closes #597.

## The problem

`-j` defaults the node count to the terminal height, so JSON output is silently truncated to whatever happens to fit on screen:

```
# a tree of 80 directories + 80 files
dust -j            ->  21 nodes
dust -j -n 100000  -> 161 nodes
```

Nothing warns you. In a CI pipeline there's no terminal at all, and callers end up passing arbitrary magic numbers like `-n 10000000` to get a complete tree.

Terminal height is a rendering constraint; it has no meaning for a machine-readable format.

## The fix

When the output is JSON, default the node count to unlimited. This mirrors the branch immediately below it, which already does exactly this for an explicit `--depth`:

```rust
if depth != usize::MAX || config.get_output_json(&options) { usize::MAX } else { ... }
```

An explicit `-n` still wins, whether it comes from the flag or the config file — `number_of_lines` is an `Option` with no clap default, so "user asked for N" and "nobody asked" are already distinguishable. Verified:

```
dust -j                                  -> 161 nodes
dust -j -n 5                             ->   6 nodes
config file number-of-lines = 5, dust -j ->   6 nodes
dust (non-JSON)                          ->  21 lines, byte-identical to before
```

`-d`, `-b` and piped output are unaffected.

## Scope

The issue floats two proposals. This is the first one, **narrowed to the node count only** — depth still defaults exactly as before, and the `--no-limit` flag from proposal 2 isn't here either.

That narrowing matters for reading the thread: you replied *"I suppose that is reasonable. Why am I constraining the json output to the height of the terminal. ?"*, and @Malix-Labs then said *"I don't think it should be done by default for sure"* — prefaced by not having looked at the codebase. As I read it, that caveat lands on unlimited *depth* (which can explode on a large filesystem) rather than on the node count, which is why I left depth alone. Quoting both so you can judge rather than take my reading of it.

## Tests

One test added to `config.rs`'s existing test module, covering JSON-on, explicit-`-n`-wins, and JSON-off.

```
cargo build:        clean
cargo test:         31 + 8 + 31 + 4 passed, 0 failed
cargo fmt --check:  clean
cargo clippy:       2 warnings, both pre-existing in tests/
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -467,4 +467,37 @@ mod tests {
         let args = get_args(vec!["dust", "--number-of-lines", "5"]);
         assert_eq!(c.get_number_of_lines(&args), Some(5));
     }
+
+    #[test]
+    fn test_get_number_of_lines_with_output_json() {
+        // Json output and no number-of-lines: main defaults this to usize::MAX.
+        let c = Config::default();
+        let args = get_args(vec!["dust", "--output-json"]);
+        assert!(c.get_output_json(&args));
+        assert_eq!(c.get_number_of_lines(&args), None);
+
+        // Json output from config and no number-of-lines.
+        let c = Config {
+            output_json: Some(true),
+            ..Default::default()
+        };
+        let args = get_args(vec![]);
+        assert!(c.get_output_json(&args));
+        assert_eq!(c.get_number_of_lines(&args), None);
+
+        // An explicit number-of-lines still wins over the json default.
+        let c = Config::default();
+        let args = get_args(vec!["dust", "--output-json", "--number-of-lines", "5"]);
+        assert!(c.get_output_json(&args));
+        assert_eq!(c.get_number_of_lines(&args), Some(5));
+
+        // A number-of-lines from the config file also wins.
+        let c = Config {
+            number_of_lines: Some(3),
+            ..Default::default()
+        };
+        let args = get_args(vec!["dust", "--output-json"]);
+        assert!(c.get_output_json(&args));
+        assert_eq!(c.get_number_of_lines(&args), Some(3));
+    }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -154,13 +154,14 @@ fn main() {
 
     let depth = config.get_depth(&options);
 
-    // If depth is set, then we set the default number_of_lines to be max
+    // If depth is set, or the output is json (which is not rendered to a
+    // terminal), then we set the default number_of_lines to be max
     // instead of screen height
 
     let number_of_lines = match config.get_number_of_lines(&options) {
         Some(val) => val,
         None => {
-            if depth != usize::MAX {
+            if depth != usize::MAX || config.get_output_json(&options) {
                 usize::MAX
             } else {
                 get_height_of_terminal()


Malix-Labs:
That's an improvement for json for sure but I don't even think that the "limiting to screen height" is a sane default anyway even for non-json

It's good to have it when you want to restrict to screen height, but that's non-deterministic and surprising, it should at least require an explicit flag

VXNCXNX:
Agreed that it's surprising, and the non-determinism is the part I'd find hardest to defend — the same command gives different output depending on how tall your terminal happens to be, and piping to a file gives different output again.

That said, I'd rather not fold it into this PR. Changing the default for the TUI path alters what every existing user sees on every run, which is a different order of change from fixing a machine-readable format, and it's @bootandy's call rather than something to slip in alongside. This PR stays at "JSON is not a TUI", which is the part he already said he agreed with.

If he wants the non-JSON default changed too, I'm happy to send that as a separate PR — it's a small change on top of this one, and it deserves its own discussion where he can weigh the compatibility cost.

Malix-Labs:
You could submit another if you want yes, [stacked](https://docs.github.com/en/pull-requests/reference/stacked-prs-cli-commands) on top of this one

VXNCXNX:
Noted, thanks. I'd rather wait for @bootandy to say he wants the non-JSON default changed before writing it, since that one alters what every existing user sees on every run. If he says yes, I'll send it stacked on this.

Malix-Labs:
It's a PR anyway, it doesn't need a green light to be submitted, just a green light to be merged

bootandy:
>the same command gives different output depending on how tall your terminal happens to be

Yes this is explicitly the goal of this tool. It's meant to fill your terminal with useful data not be piped into other commands. Obviously it can be overridden with either -d or -n. 

>The non-JSON default changed 

No please don't change this. 


However, in the case of JSON - I agree that defaulting to the terminal height is not correct. 

What I am open to is do we pick an arbitrary number (~30) and default to that. Or we assume they want everything. I lean towards everything. 

Thanks

VXNCXNX:
That is what this PR does, and only that. Sorry, the description talked about the non-JSON default enough to make it look like I was proposing to change it. I am not.

The whole change is one condition in `main.rs`:

```rust
-            if depth != usize::MAX {
+            if depth != usize::MAX || config.get_output_json(&options) {
```

So JSON gets everything, which is the option you lean towards, and terminal output keeps filling the terminal exactly as before. `-n` and `-d` still win over the JSON default, from the flag or the config file, which the tests pin.

The paragraph about terminal height was me explaining why JSON is different, not arguing the general case. Your reasoning for the terminal default makes sense and it stays untouched here.

bootandy:
yes - just tested locally looks good, merging.

PR #600, opened 2026-08-14
feat: add mount point exclusion

wakamex:
## Why

I use `dust` to list immediate project sizes before moving directories between disks, and I want mounted directories omitted explicitly.

For my current tree, `-x` produces the same result because the mounted child is on another filesystem. But `-x` checks filesystem device metadata throughout the walk, while an explicit mount-point filter can skip known mount targets directly. It also covers same-filesystem bind mounts, which device-ID filtering cannot distinguish from ordinary directories.

Paired local timings with byte-identical JSON output:

- `dust -d 1 -D -x`: 28.7 seconds
- `dust -d 1 -D --exclude-mounts`: 20.9 seconds

## What changed

`--exclude-mounts` skips mount points below each supplied path while still allowing a supplied path that is itself mounted.

On Linux it reads `/proc/self/mountinfo`, which covers same-device bind mounts. Other platforms use the existing `sysinfo` disk mount-point inventory. The option is also available as `exclude-mounts=true` in the config file.

The README, sample config, man page, and generated shell completions are updated.

## Validation

- `cargo fmt --all -- --check`
- `cargo test --all-targets` - 80 tests passed
- `cargo clippy -- -D warnings`
- `cargo check --target x86_64-pc-windows-gnu`
- Isolated Linux test with source and bind target on the same device

OpenAI Codex assisted with code generation, testing, and review. I reviewed and validated the final patch, and the commit includes an `Assisted-by` trailer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ Usage: dust -r (reverse order of output)
 Usage: dust -o si/b/kb/kib/mb/mib/gb/gib (si - prints sizes in powers of 1000. Others print size in that format).
 Usage: dust -X ignore  (ignore all files and directories with the name 'ignore')
 Usage: dust -x (Only show directories on the same filesystem)
+Usage: dust --exclude-mounts (Do not descend into mount points below the supplied paths)
 Usage: dust -b (Do not show percentages or draw ASCII bars)
 Usage: dust -B (--bars-on-right - Percent bars moved to right side of screen)
 Usage: dust -i (Do not show hidden files)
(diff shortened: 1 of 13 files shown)


wakamex:
After checking my actual mount layout more carefully, `-x` already has the semantics I need because the only descendant mount is on a different filesystem. The performance gap that led me here comes from redundant metadata work in the current `-x` path, and #575 already addresses that. Closing this in favor of #575. Thanks.

PR #602, opened 2026-08-15
deps: bump crossbeam-epoch to 0.9.20 for RUSTSEC-2026-0204

### LINKED ISSUE #601: cargo audit reports RUSTSEC-2026-0204 (crossbeam-epoch 0.9.18) via rayon dependency
Title: `cargo audit` reports RUSTSEC-2026-0204 (crossbeam-epoch 0.9.18) via rayon dependency

## Summary

`cargo audit` flags one vulnerability in the current `Cargo.lock`:

```
Crate:     crossbeam-epoch
Version:   0.9.18
Title:     Invalid pointer dereference in `fmt::Pointer` impl for `Atomic` and `Shared`
           when the underlying pointer is invalid
Date:      2026-07-06
ID:        RUSTSEC-2026-0204
URL:       https://rustsec.org/advisories/RUSTSEC-2026-0204
Solution:  Upgrade to >=0.9.20
```

Dependency path:

```
crossbeam-epoch 0.9.18
└── crossbeam-deque 0.8.6
    └── rayon-core 1.13.0
        └── rayon 1.11.0
            └── du-dust (this crate)
```

`rayon` is a direct dependency (`Cargo.toml`: `rayon = "1"`) used in `src/dir_walker.rs` for parallel directory traversal.

## Root cause / how to trigger

Affected versions (`>=0.9.0, <0.9.20`) of `crossbeam-epoch`'s `fmt::Pointer` (`{:p}`) implementation for `Atomic<T>` / `Shared<T>` dereference the pointer to build the formatted output, instead of just printing the address. When the pointer is null (e.g. `Atomic::null()` / `Shared::null()`), this dereferences a null pointer. Fixed in [crossbeam-rs/crossbeam#1276](https://github.com/crossbeam-rs/crossbeam/pull/1276), released in `crossbeam-epoch` 0.9.20.

Minimal reproduction (outside of this repo, standalone crate depending directly on `crossbeam-epoch = "=0.9.18"`):

```rust
use crossbeam_epoch::Atomic;

fn main() {
    let a: Atomic<u64> = Atomic::null();
    println!("{:p}", a); // panics on 0.9.18, prints "0x0" on 0.9.20+
}
```

On 0.9.18 this aborts with:

```
thread 'main' panicked at .../crossbeam-epoch-0.9.18/src/atomic.rs:204:9:
null pointer dereference occurred
thread caused non-unwinding panic. aborting.
```

On 0.9.20 it prints `0x0` and returns normally.

## Impact on du-dust specifically

I checked whether this is actually reachable from du-dust's own code paths: neither `crossbeam-deque` 0.8.6 nor `rayon-core` 1.13.0 ever format an `Atomic`/`Shared` value with `{:p}`/`fmt::Pointer` anywhere in their source, and du-dust does not depend on `crossbeam-epoch` directly. So in practice this advisory is not exploitable through du-dust's normal usage — it's purely a transitive `Cargo.lock` entry that trips `cargo audit`, which matters for downstream packagers/CI that gate on a clean audit (similar to #324).

## Fix

Bumping the lockfile entry is sufficient — no source changes needed:

```
cargo update -p crossbeam-epoch --precise 0.9.20
```

(`rayon-core`'s `crossbeam-deque` requirement already accepts `crossbeam-epoch` `^0.9`, so this resolves cleanly.) Re-running `cargo audit` afterward reports 0 vulnerabilities.

Could a maintainer run `cargo update -p crossbeam-epoch` and cut a patch release so downstream package maintainers (Homebrew, distro repos, etc.) picking up `Cargo.lock` get the clean audit? Happy to open a PR with the updated lockfile if useful.

VXNCXNX:
Fixes #601.

## What's broken

`Cargo.lock` pins `crossbeam-epoch 0.9.18`, which RUSTSEC-2026-0204 covers. It is a transitive dependency, rayon 1.11 to rayon-core 1.13 to crossbeam-deque 0.8.6.

The advisory is a null pointer dereference in the `fmt::Pointer` impl for `Atomic` and `Shared`, patched in 0.9.20.

## The fix

`cargo update -p crossbeam-epoch`. Two lines in the lockfile, nothing in `Cargo.toml`, since it already declares `rayon = "1"` and the bump is semver compatible. dust ships a binary, so the lockfile is the right place for this.

## Verification

```
Updating crossbeam-epoch v0.9.18 -> v0.9.20
```

`cargo build --release` succeeds and the whole test suite passes: 32, 8, 31 and 4 tests across the four targets, zero failures.

One thing to note that is not from this change: `cargo clippy --all-targets -- -D warnings` currently fails on master with four `empty_line_after_doc_comments` errors in the test files, a lint that Rust 1.97 turned on. I confirmed those are present on a clean tree with my change stashed, so they are unrelated to this PR, but CI may be red until someone fixes them.

I did not run `cargo audit` myself, it is not installed here. I verified the pinned version and the advisory's patched range directly instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -252,9 +252,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b82ac4a3c2ca9c3460964f020e1402edd5753411d7737aa39c3714ad1b5420e"
+checksum = "2d6914041f254d6e9176c01941b21115dcfb7089e55135a35411081bd106ef3f"
 dependencies = [
  "crossbeam-utils",
 ]


bootandy:
Good work, but I did a much needed cargo update on everything instead here: https://github.com/bootandy/dust/pull/604

sorry I've been offline for a while and am catching up on maintenance.

PR #603, opened 2026-08-15
fix: don't panic on an invalid --mtime/--atime/--ctime value

VXNCXNX:
## What's broken

A non-numeric or empty value for `--mtime`, `--atime` or `--ctime` aborts the process.

```
$ dust --mtime "abc" .
thread 'main' panicked at src/config.rs:220:41: invalid data format

$ dust --mtime "" .
same panic, from the .expect on chars().next()
```

## The fix

`get_filter_time_operator` used `unwrap_or_else(|_| panic!(...))` on the parse and `.expect(...)` on the first character. Both now warn on stderr and return `None`, which is already the supported no-filter state: `utils.rs:95` matches `None => false`, so nothing gets excluded.

This follows `convert_min_size` a few lines below in the same file, which already does exactly this for a bad `--min-size`.

```
$ dust --mtime "abc" .
Ignoring invalid time filter: abc
(normal output)
```

Valid values are untouched. The parse and the arithmetic are unchanged, only hoisted above the match, so `+5`, `-5` and `5` still produce the same operator and timestamp. Rust accepts the leading `+`, and `-5` still parses negative and gets `.abs()`ed as before.

## Verification

Two tests in the existing `mod tests`. One covers the empty and non-numeric cases, the other pins the exact `Operator` and time for `+5`, `-5` and `5` so a future change to the hoisting cannot silently alter them. With the `panic!` restored the first test fails on the panic itself. `cargo test` passes, 77 tests across the suite.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -214,13 +214,16 @@ fn get_filter_time_operator(
 ) -> Option<(Operator, i64)> {
     match option_value {
         Some(val) => {
-            let time = current_date_epoch_seconds
-                - val
-                    .parse::<i64>()
-                    .unwrap_or_else(|_| panic!("invalid data format"))
-                    .abs()
-                    * DAY_SECONDS;
-            match val.chars().next().expect("Value should not be empty") {
+            let Some(first_char) = val.chars().next() else {
+                eprintln!("Ignoring invalid time filter: {val}");
+                return None;
+            };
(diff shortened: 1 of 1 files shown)


bootandy:
I don't want to ignore an invalid time - I want to fail if it is wrong.

PR #606, opened 2026-08-15
fix: fail cleanly on an invalid --mtime/--atime/--ctime instead of panicking

VXNCXNX:
Follow-up to #603, which you closed with:

> I don't want to ignore an invalid time - I want to fail if it is wrong.

That is the right call and this does that instead.

## What's broken

An unparseable `--mtime`, `--atime` or `--ctime` value aborts with a panic.

```
$ dust --mtime "abc" .
thread 'main' panicked at src/config.rs:220:41: invalid data format

$ dust --mtime "" .
same panic, from the .expect on chars().next()
```

## The fix

Both sites now print the reason on stderr and exit 1, rather than panicking or, as #603 proposed, carrying on:

```
$ dust --mtime "abc" . ; echo $?
Invalid value for time filter: "abc"
1

$ dust --mtime "" . ; echo $?
Invalid value for time filter: ""
1
```

This follows the `Regex::new` handling in `main.rs`, which already does exactly this for a bad `-e` value: message on stderr, `process::exit(1)`, no traceback.

Valid values are untouched. The parse and the arithmetic are unchanged, only hoisted above the match, so `+5`, `-5` and `5` still produce the same operator and timestamp, and still exit 0.

The second guard, on `chars().next()`, is unreachable now that the parse runs first: only an empty string has no first character, and `"".parse::<i64>()` already fails. It is left in place with a comment rather than unwrapped, since removing it would mean an unchecked `unwrap` on a value the compiler cannot prove non-empty.

## Verification

`test_invalid_time_filter_fails_cleanly` in `tests/test_flags.rs`, covering all three flags against both the non-numeric and the empty value. It runs the real binary through `assert_cmd`, so it observes the actual exit status, and it asserts three things: non-zero exit, the reason present on stderr, and the word `panicked` absent.

With the change reverted it fails on the panic itself. `cargo test` passes, 75 tests across the suite, and `cargo fmt -- --check` and `cargo clippy --all-targets` introduce nothing new.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -5,6 +5,7 @@ use regex::Regex;
 use serde::Deserialize;
 use std::path::Path;
 use std::path::PathBuf;
+use std::process;
 
 use crate::cli::Cli;
 use crate::dir_walker::Operator;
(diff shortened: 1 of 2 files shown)


bootandy:
you sound like a claude instance. Are you?

bootandy:
we probably dont need the tests, but yes I suppose this version is better than the current one

VXNCXNX:
Yes. I use Claude to draft these. I read the diff and run the tests before anything goes out, and I answer review comments myself, but you should know a model wrote the first version of the patch and the description.

Happy to say so on future PRs here, or to stop sending them, whichever you prefer. Your call and no hard feelings either way.

Tests dropped, the PR is just the `config.rs` change now.

bootandy:
that looks good thank you

PR #608, opened 2026-08-15
Fix Windows file time conversion (from FILETIME to unix time)

adithya-n05:
Hi Andy! I'm Adithya, I met you on Thursday earlier this week

I wanted to practice contributing to some open source projects like you suggested, and I thought I'd start by contributing to `dust`!

This PR fixes the following:
- For windows machines, dust reads time values in [platform.rs](https://github.com/bootandy/dust/blob/03428869ab126665c13f1222509e4039ffa35187/src/platform.rs#L149-L151) as FILETIME values. Those are 100 nanosecond intervals since 1601 (i.e. not unix timestamps)
- These will be later passed to chrono by calls to `get_pretty_file_modified_time` in [display.rs](https://github.com/bootandy/dust/blob/03428869ab126665c13f1222509e4039ffa35187/src/display.rs#L380) and here chrono will expect unix time

Correction:
- Added a conversion function to convert filetime to unix timestamps if on a windows machine, and tests. Replaced all timestamp reads to call this function.

### DIFF
--- a/src/platform.rs
+++ b/src/platform.rs
@@ -13,6 +13,14 @@ fn get_block_size() -> u64 {
 type InodeAndDevice = (u64, u64);
 type FileTime = (i64, i64, i64);
 
+#[cfg(target_family = "windows")]
+fn filetime_to_unix_seconds(filetime: u64) -> i64 {
+    const TICKS_PER_SECOND: i128 = 10_000_000;
+    const UNIX_EPOCH_FILETIME: i128 = 116_444_736_000_000_000;
+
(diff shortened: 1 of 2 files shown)


bootandy:
Hi yes I remember you. 

In advance I'm not reliably online so I apologise in advance if I don't reply. 

It would be good to have someone looking after the windows build - because I don't have a windows box myself.

bootandy:
i think your reasoning is ok but the tests currently fail. I think the copilot review comments are worth looking at.

adithya-n05:
@bootandy I've fixed the failing tests and addressed the copilot feedback

When you may have a chance, could you please approve the workflow so CI can run again?

PR #609, opened 2026-08-16
fix: --collapse on the command line is ignored when the config sets it

VXNCXNX:
`--collapse` on the command line is ignored whenever the config file also sets
`collapse`. The config wins, which is the opposite of how every other option
behaves.

Tree `/tmp/dusttest` with `node-modules/` and `src/`, config file with
`collapse=["src"]`, run as
`dust -P -c --config /tmp/dustconf.toml --collapse=node-modules .`:

```
before: (the flag is ignored, src is collapsed and node-modules is expanded)
 24Ki   ┌── src
 52Ki   │     ┌── b.bin
 56Ki   │   ┌─┴ deeper
100Ki   │   ├── a.bin
160Ki   │ ┌─┴ deep
164Ki   ├─┴ node-modules
192Ki ┌─┴ .

after:  (the flag wins)
 20Ki     ┌── main.rs
 24Ki   ┌─┴ src
164Ki   ├── node-modules
192Ki ┌─┴ .
```

## Cause

`Config::get_collapse` branches on the config value instead of the flag:

```rust
if self.collapse.is_none() {
    options.collapse.clone()
} else {
    self.collapse.clone()
}
```

So the config is returned whenever it is set, and the flag is only consulted
when it is not. `get_threads`, `get_number_of_lines` and `get_custom_stack_size`
all do the reverse, testing the command-line value first.

`collapse` was added to the config in #552, and this branch has been inverted
since.

## The fix

Test `options.collapse` first, matching the other three. A config-only setup is
unchanged, verified separately.

## Verification

`test_get_collapse` covers all four combinations: neither set, flag only, config
only, and both.

Reverting only the branch order fails it:

```
thread 'config::tests::test_get_collapse' panicked at src/config.rs:429:9:
assertion `left == right` failed
  left: Some([".git"])
 right: Some(["node_modules"])
```

The config-only and flag-only cases still pass under that mutation, which is
what isolates the both-set case as the real bug.

`cargo test` is 8 + 31 + 4 + 32 + 0 passed, 0 failed. `cargo fmt --check` clean.

One note: `cargo clippy --all-targets -- -D warnings` reports two
`empty line after doc comment` errors in `tests/test_exact_output.rs:10` and
`tests/test_flags.rs:5`. Both are present on master before this change and are
in files I did not touch, so I left them alone.

*Disclosure: written with AI assistance (Claude Code). I built binaries before and after, produced the trees above by running them, and ran the mutation check myself.*

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -188,10 +188,11 @@ impl Config {
     }
 
     pub fn get_collapse(&self, options: &Cli) -> Option<Vec<String>> {
-        if self.collapse.is_none() {
-            options.collapse.clone()
-        } else {
+        // command line wins, as in get_threads and get_number_of_lines
+        if options.collapse.is_none() {
             self.collapse.clone()
+        } else {
+            options.collapse.clone()
         }
     }
 }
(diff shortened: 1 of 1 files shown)


bootandy:
that is a fair point. I don't think we need the tests though.

VXNCXNX:
Done, test dropped. The fix is the one-line branch order in `get_collapse`.

For the record, since the mutation proof lived in the test I just removed: reverting only the branch order made it return the config value where the flag was set, which is the whole bug. The before and after trees in the description came from running the binaries, so that part still stands on its own.

PR #619, opened 2026-09-08
Show partial scan results on Ctrl-C

h0tmi:
Pressing Ctrl-C during a long scan currently prints `Aborting` and exits, discarding the statistics already collected. This change stops traversal and renders the collected entries through the usual aggregation and tree display, retaining partially scanned directories and their ancestors.

- The first Ctrl-C displays a partial result; a second Ctrl-C exits immediately. Interrupts before scanning also exit immediately.
- Text output includes `Partial result: scan interrupted by Ctrl-C.` JSON output adds `"partial": true` at the root while preserving the format of completed results.
- Interrupted runs exit with status 130. Sizes and percentages reflect only the scanned entries.
- The README documents the new behavior.

Validation on macOS (Apple Silicon):

- `cargo test --locked --offline` — all 84 tests passed.
- `cargo fmt --all -- --check` — passed.
- `cargo clippy --locked --offline --bin dust --test test_interrupt -- -D warnings` — passed.
- Release build and installed binary smoke test — passed.

The new Unix integration tests send SIGINT after observing scan progress and verify the partial tree, valid JSON, retained files, aggregate counts, and exit status. Unit tests cover cancellation before starting another path and stopping interrupted I/O retries; a regression test checks that completed JSON retains its existing shape.


shotInLeg:
llm slop?

src/dir_walker.rs
```diff
@@ -355,9 +372,33 @@ mod tests {
             follow_links: false,
             progress_data: indicator.data.clone(),
             errors: Arc::new(Mutex::new(RuntimeErrors::default())),
+            interrupted: Arc::new(AtomicBool::new(false)),
         }
     }
 
+    #[test]
```
shotInLeg, line 379: Move it to file for tests, no?

h0tmi: Image

tests/test_flags.rs
```diff
@@ -24,6 +24,16 @@ fn build_command<T: AsRef<OsStr>>(command_args: Vec<T>) -> String {
     str::from_utf8(&finished.stdout).unwrap().into()
 }
 
+#[test]
```
shotInLeg, line 27: Test in test file! - you can if you want

bootandy:
Sorry, I'm not keen on this. 

If the data is partially collected we don't know how to build the size bars. So I'd rather it cancelled and showed nothing

PR #620, opened 2026-09-09
perf(walk): flatten recursion via rayon::scope, drop -S stack-size knob

bootandy:
Assisted-by: Claude Opus 4.7 (code generation, refactoring, code review)


bootandy:
This does have a performance impact (aprox double) so is worth doing.

I hope the old issue with low memory on small machines is now fixed as we don't need a large stack.
https://github.com/bootandy/dust/issues/256

PR #622, opened 2026-09-09
ci: remove i686-pc-windows-gnu

bootandy:
build was failing:
rustup toolchain install stable-i686-pc-windows-gnu Run rustup toolchain install stable-i686-pc-windows-gnu --target i686-pc-windows-gnu --profile minimal --no-self-update
  error: toolchain 'stable-i686-pc-windows-gnu' may not be able to run on this system
  note: to build software for that platform, try `rustup target add i686-pc-windows-gnu` instead
  note: add the `--force-non-host` flag to install the toolchain anyway
  Error: Process completed with exit code 1.

an i686 windows gnu is a weird combination, so I don't think we'll miss this


mati865:
All i686 Windows host tools are phased out on nightly. To produce 32-bit Windows binaries, be it i686-pc-windows-gnu, i686-pc-windows-gnullvm, or i686-pc-windows-msvc, you must cross-compile them from a supported host toolchain.

